PR #11, opened 2014-05-02
Allow user to override the host header when using SPDY

### LINKED ISSUE #10: No way to override :host header when using SPDY.
When using SPDY, there is no way to connect to a certain IP and specify a different `:host` header.

betawaffle:
This fixes #10 for me.


essen:
Hello, made a few changes to request code, including this one, hopefully should be good now. Thanks!

PR #13, opened 2014-05-23
Add {require_connect, true} option

stolen:
When `{require_connect, true}` specified,  `{ok, Pid}` is returned only after connection is established.

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -98,7 +98,8 @@
 	transport :: module(),
 	protocol :: module(),
 	proto_opts :: gun_http:opts(), %% @todo Make a tuple with SPDY and WS too.
-	protocol_state :: any()
+	protocol_state :: any(),
+    ack_sent :: boolean()
 }).
 
 %% Connection.
@@ -132,6 +133,8 @@ open_opts([{retry_timeout, T}|Opts]) when is_integer(T) > 0 ->
 open_opts([{type, T}|Opts])
 		when T =:= tcp; T =:= tcp_spdy; T =:= ssl ->
 	open_opts(Opts);
+open_opts([{require_connect, RC}|Opts]) when is_boolean(RC) ->
+    open_opts(Opts);
 open_opts([Opt|_]) ->
 	{error, {options, Opt}}.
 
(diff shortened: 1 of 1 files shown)


essen:
I am thinking of sending a message on connect success/failure, not sure yet exactly how it will look like, but would that kind of message be good enough for your use case? Basically you need to wait for the connection to be up, and that's what it'll be for, except it'll also work when the connection is lost and then back up.

stolen:
Yes, for my case messages would be even better than proposed patch (as I wrote earlier the idea of asynchronous connect aknowledges came to me too late).
By the way, if no one else needs this functionality, I may stay happy with my fork and when I need connect aknowledgement with fresh gun I'll make another pull request.

essen:
OK I think it should be like this:

* connection success: send a message to say we are connected
* connection loss: send a message to say we lost the connection

On connection loss we should check whether it is possible to retry requests for which we haven't gotten a response yet. The rules for doing this are defined in the HTTP RFC (basically, idempotent is OK to retry). The connection loss message should inform the user whether everything is OK and will continue as intended or if user input is needed.

We will need an option to Websocket to indicate whether to reconnect automatically in Websocket mode when the connection comes back up. Whether this is desirable is application dependent. If the option is enabled then we automatically reconnect using the same request as before. Otherwise the connection loss message will say that we can't go back to Websocket automatically (this should be a different value than the part about idempotent methods).

The gun:open/{2,3} function should be enhanced to accept a 'sync' option that, when enabled, automatically wait for the connection success message.

Perhaps it can also be useful to add an option to send a message on each connection failure, this can be useful for debugging among other things.

essen:
The connection success message should also indicate what protocol was selected.

essen:
This is mostly done in a738db07c2040461f0fd431f04ecf90b157ead5b. However I still want to add a `sync` option. I will do that tomorrow though.

essen:
OK a sync option is a bad idea, I will add a new function that waits for the up message and allows passing a monitor (otherwise creates one).

essen:
Reason it's a bad idea is that we would end up discarding the selected protocol or making the open return value different depending on sync and I don't like that much.

essen:
Done in 6676a982ff18d6a6b346f608e7ce313cfccc873e. Thanks for the feedback!

PR #23, opened 2014-06-18
Return body when content-length not specified for extend/gun#22

unix1:
Return body when content-length not specified for extend/gun#22.
And I even capitalized the first letter of the commit message! ;)


essen:
I have changed point 2 to be separate for request and response. I will see if I the PR merges cleanly and otherwise make the small fix manually. Thanks a bunch!

essen:
Merged, thanks!

unix1:
Exciting! Thanks for separating the request/response functions - that's also what I had done in another temporary branch to make it usable for my use case. I'll subject it to testing on my side separately and open a new issue if I find anything interesting.

PR #27, opened 2014-08-22
Accept SSL options in gun:open/3's opt argument

nwt:
This makes it possible to use an SSL client certificate with gun (among other things).


essen:
Done in 8366ba94bb9e450221a246acdd482c0162affcd9. The `transport_opts` option can be used for specifying transport options. Currently the value is unrestricted but do note that it will be restricted later on. Thanks and enjoy!

PR #29, opened 2014-08-28
Include non-standard ports in host header

woahdae:
As per RFC 2616, non-standard ports must be sent with the host name in
the host header as specified by the URI.

Previously, only the host name was sent; fixed.


essen:
Hello, I made recently a change similar to this but also added the appropriate code for the SPDY part. Thanks for the patch!

PR #34, opened 2015-02-21
Update gun.md

alexr:
Fixing simple copy-paste mistake.


essen:
Merged locally, will be included in the push when I'm done updating the docs. Thanks!

PR #71, opened 2015-05-19
Handle non-zero content lengths when performing a HEAD request

srstrong:


essen:
I have done a very similar fix when I hit this a few days ago, and forgot to check existing PRs beforehand. Sorry about that. Fix is https://github.com/ninenines/gun/commit/0414eede62946f995a5ab4879f9026548440199c

Thanks!

Issue #63, opened 2015-03-28
Add Websocket protocol negotiation


essen:
Options are needed, code also, and guide of course.

mururu:
I need this. Is there any progress here?

essen:
You can set/check the headers directly currently. This ticket is for a more automated way.

mururu:
The headers seems to be ignored because the `Headers` here is not merged.

https://github.com/ninenines/gun/blob/master/src/gun_http.erl#L391-L416

essen:
Good catch!

essen:
Pushed a different Should be good on current master now. Thanks!

I do not think I will add more with regard to Websocket, since we can both send and receive headers now. Closing, thanks!

PR #72, opened 2015-06-08
Fix ws_upgrade to merge argument headers

seizans:
As mentioned at https://github.com/ninenines/gun/issues/63 , I merged `Headers` argument into `Headers2` to be able to set headers.


essen:
Looks good, thanks. Will merge after EUC.

mururu:
:+1:

voluntas:
:+1:

essen:
Pushed a different fix since I want to avoid ++. Should be good on current master now. Thanks!

seizans:
:+1:

PR #76, opened 2015-07-18
Handle 204 regardless of content length

petrohi:


essen:
Thanks, now that I am finally settled in and have Internets I will be able to merge this, hopefully before the week ends. :-)

essen:
The way I see it, you have to check 204 *before* checking content-length, otherwise a sneaky invalid 204 with content-length header will break everything. Will make a patch soon.

essen:
Pushed the fix I had in mind. Ping if still an issue. :-)

PR #77, opened 2015-07-18
Pass headers in ws_upgrade

petrohi:


essen:
I think there's another PR for this. Will get back to you.

essen:
Pushed a different fix since I want to avoid ++. Should be good on current master now. Thanks!

PR #78, opened 2015-07-22
Fix ssl:negotiated_next_protocol/1 is deprecated for OTP 18.

mkrentovskiy:


essen:
Don't remember if we support only 18 yet or not.

essen:
Merged, thanks!

Wanted to try the hub tool to deal with PRs. Simpler. :-)

PR #83, opened 2015-08-31
Use https instead of git for ranch and cowlib dependencies

### LINKED ISSUE #82: Use https instead of git for ranch and cowlib dependencies
This would be compatible with Cowboy, and not require GitHub authentication to fetch them.

rightfold:
This is compatible with Cowboy and doesn't require authentication with
GitHub. Rebar and Mix do not automatically turn git:// into https://.

Fixes #82


essen:
Thanks will merge soon, and don't worry about the build failure, it's all expected errors.

essen:
Merged, thanks!

PR #85, opened 2015-09-15
Honor retry => 0 option even when host is down initially.

MSch:
Previously when starting a gun connection when the host is down the
connection would never give up retrying to connect if `retry`
was set to 0.

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -489,7 +489,7 @@ connect(State=#state{host=Host, port=Port, opts=Opts, transport=Transport=ranch_
 			end,
 			up(State, Socket, Protocol, ProtoOptsKey);
 		{error, _} ->
-			retry(State, Retries - 1)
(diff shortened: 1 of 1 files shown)


essen:
Looking in my local repo it looks like I had already worked on this, but for some reason didn't commit. I suspect I didn't because I didn't have a test. Can you provide one (even just something manual in the shell)?

ppolv:
Got the same behavior,  the issue is that after each failed connection attempt (including the first one),   gun decrements the Retry count by one, and then tries to performs a retry.
So,  if Retries was initially 0 this never match and the counter keeps going negative:
``  retry(_, 0) ->   
ok;
``
On my case I just passed retry =>  1   that has the desired effect (no retry is attempted)

essen:
This is fixed in the most recent commit, and I've added a test suite to confirm the behavior is as intended. Thanks!

PR #87, opened 2015-10-22
fixed typo in code snippet

magicgoose:
…I thought it would be better if the code can be compiled. Currently there's syntax error (`;` after last match clause)


essen:
It was on purpose but you're right, it makes more sense like this.

essen:
Merged, thanks!

PR #94, opened 2016-01-05
Updated gun.app.src with metadata for hexer app to be uploaded on hex.pm

davecaos:
Hello Loïc. Please, could you add this metadata to gun? it is necessary to run [hexer](https://github.com/inaka/hexer) for upload the project to hex.pm .
Thanks.


essen:
No because the .app.src file will be removed in a future update. Even if it wasn't, some tools produce warnings on incorrect values (reltool I believe) and we don't want that.

PR #98, opened 2016-06-06
Fix badarg in http/2 ping generation

philipstears:
cow_http2:ping/1 wants an integer, not a binary.

Originally I patched cow_lib to take a binary (ninenines/cowlib#38), but as @essen pointed out, parsing produces an integer, so instead, I've changed gun to send an integer.


essen:
Merged, thanks!

PR #103, opened 2016-09-22
Allow the client to know the underling reason of the disconnection

ppolv:
The user of gun might need to do different things depending
on the specific cause of the error. Some are transient while others
are permanent, or require specific actions.  For example expired
client certificates.

The underling reason of the disconnection was never propagated to the
client, this patch fixes that. In case there was retries configured,
the error returned is the one on the last retry attempted.

Note:
  This might break other people code if they were relying on the
  general 'gone' cause that was always given so far
  (it was not part of the api definition nor documented anyway).


essen:
Good idea but should be wrapped in a tuple like `{gone, Reason}` because `gone` indicates that Gun stopped trying to reconnect.

essen:
I have just pushed a simpler version of this patch with a test. Thanks!

Issue #106, opened 2016-11-03
no connect timeout


sugus-seu:
hi, I notice that there's no timeout option in connect stage. When init connect, we can count on gen_server init timeout, but when retry connect, there's no timeout, so the whole process hang on.

essen:
Good point. Patch welcome!

essen:
Done, thanks!

PR #107, opened 2016-11-04
aad connet timeout

sugus-seu:
[no connect timeout #106](https://github.com/ninenines/gun/issues/106)

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -16,7 +16,7 @@
 
 %% Connection.
 -export([open/2]).
--export([open/3]).
+-export([open/4]).
 -export([close/1]).
 -export([shutdown/1]).
 
@@ -68,8 +68,8 @@
 -export([dbg_send_raw/2]).
 
 %% Internals.
--export([start_link/4]).
--export([init/5]).
+-export([start_link/5]).
+-export([init/6]).
 -export([system_continue/3]).
 -export([system_terminate/4]).
 -export([system_code_change/4]).
@@ -106,7 +106,8 @@
 	socket :: inet:socket() | ssl:sslsocket(),
 	transport :: module(),
 	protocol :: module(),
-	protocol_state :: any()
+	protocol_state :: any(),
+    timeout :: timeout()
 }).
 
 %% Connection.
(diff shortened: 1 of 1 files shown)


essen:
Use Opts, don't add an argument.

sugus-seu:
I use an argument because ranch did that way, the opts are more like options of net layer

sugus-seu:
I use an argument because ranch did that way, the opts are more like options of net layer

sugus-seu:
I use an argument because ranch did that way, the opts are more like options of net layer

sugus-seu:
I use an argument because ranch did that way, the opts are more like options of net layer

sugus-seu:
I use an argument because ranch did that way, the opts are more like options of net layer

essen:
The options are not transport layer options. Use Opts.

sugus-seu:
done

src/gun.erl
```diff
@@ -473,7 +475,7 @@ dbg_send_raw(ServerPid, Data) ->
 
 start_link(Owner, Host, Port, Opts) ->
 	proc_lib:start_link(?MODULE, init,
-		[self(), Owner, Host, Port, Opts]).
+        [self(), Owner, Host, Port, Opts], maps:get(conn_timeout, Opts, 1000)).
```
essen: Don't really understand this part. We init_ack right at the start so this doesn't actually do anything. The start_link returns before we even try to connect.

src/gun.erl
```diff
@@ -497,7 +499,7 @@ connect(State=#state{host=Host, port=Port, opts=Opts, transport=Transport=ranch_
 		{alpn_advertised_protocols, Protocols},
 		{client_preferred_next_protocols, {client, Protocols, <<"http/1.1">>}}
 		|maps:get(transport_opts, Opts, [])],
-	case Transport:connect(Host, Port, TransportOpts) of
+    case Transport:connect(Host, Port, TransportOpts, maps:get(conn_timeout, Opts, 1000)) of
```
essen: Default of 1000 is too small. 5000 is usually good.

src/gun.erl
```diff
@@ -162,9 +162,11 @@ check_options([Opt = {protocols, L}|Opts]) when is_list(L) ->
 		_ ->
 			{error, {options, Opt}}
 	end;
+check_options([{conn_timeout, T}|Opts]) when is_integer(T), T > 0 ->
+    check_options(Opts);
 check_options([{retry, R}|Opts]) when is_integer(R), R >= 0 ->
 	check_options(Opts);
-check_options([{retry_timeout, T}|Opts]) when is_integer(T) > 0 ->
+check_options([{retry_timeout, T}|Opts]) when is_integer(T), T > 0 ->
```
essen: This is unrelated and should be in its own commit.

sugus-seu:
commit again

essen:
Merged with some changes, tests and docs; the option will be named `connect_timeout` and default to infinity. Thanks!

Issue #108, opened 2016-11-16
Add support for choosing a process to reply to


manifest:
The proposal is to distinguish the connection owner and the process that we allow to perform requests and receive stream-related messages (including response). Currently we reject any requests from processes that don't own the connection. It looks as a better approach to store pid of the process that made a request, and redirect all stream-related messages to it.

To achieve that, we could extend a [stream data type](https://github.com/ninenines/gun/blob/0414eede62946f995a5ab4879f9026548440199c/src/gun_http.erl#L42), adding a new `ReplyTo` field:

```erlang
streams = [] :: [{reference() | websocket_info(), binary(), boolean(), ReplyTo :: pid()}],
```

We could get its value from [request](https://github.com/ninenines/gun/blob/0414eede62946f995a5ab4879f9026548440199c/src/gun.erl#L610-L617):

```erlang
{request, ReplyTo, StreamRef, Method, Path, Headers} ->
	ProtoState2 = Protocol:request(ProtoState,
		StreamRef, Method, Host, Port, Path, Headers, ReplyTo),
	loop(State#state{protocol_state=ProtoState2});
{request, ReplyTo, StreamRef, Method, Path, Headers, Body} ->
	ProtoState2 = Protocol:request(ProtoState,
		StreamRef, Method, Host, Port, Path, Headers, Body, ReplyTo),
	loop(State#state{protocol_state=ProtoState2});
```

And could use it to [send stream-related messages](https://github.com/ninenines/gun/blob/0414eede62946f995a5ab4879f9026548440199c/src/gun_http.erl#L188): `gun_response`, `gun_data`, `gun_push`, `gun_error`

```erlang
send_data_if_alive(Data, #http_state{streams=[{StreamRef, _, true, ReplyTo}|_]}, IsFin) ->
	ReplyTo ! {gun_data, self(), StreamRef, IsFin, Data},
```

A real use case, HTTP connection pool where we have one process holding connection and many processes that want to perform requests.

essen:
Sounds reasonable. PR welcome. Needs a test though (if current test suites are not really helping, a manual test case would also do).

manifest:
Nice. Will do.

essen:
It might be good to have two modes, one where the behavior is more or less the same (one process is allowed to make requests), and one where the behavior is like you propose.

There's probably also some thoughts to be given with regard to Websocket.

manifest:
Do we really need to restrict requesting processes? What the use case?

[...]

PR #109, opened 2016-11-17
Add support for choosing a process to reply to

manifest:
Aims to [this](https://github.com/ninenines/gun/issues/108) issue.

There is [one case](https://github.com/ninenines/gun/blob/0414eede62946f995a5ab4879f9026548440199c/src/gun_http2.erl?ts=2#L86) when [Owner](https://github.com/ninenines/gun/blob/0414eede62946f995a5ab4879f9026548440199c/src/gun_http2.erl?ts=2#L276) still could receive an `gun_error` message for HTTP2. Because we can't find a particular stream when `connection_error` is occurred, and `StreamID` has't [parsed](https://github.com/ninenines/cowlib/blob/8e6d0f462850a90bd8a60995f52027839288d038/src/cow_http2.erl#L78-L234) yet). The same is for [SPDY](https://github.com/ninenines/gun/blob/0414eede62946f995a5ab4879f9026548440199c/src/gun_spdy.erl?ts=2#L182). I've added a note relating to this behavior to the sources.

As for the tests, it looks like `spdy_SUITE` is broken, and `twitter_SUITE` contains just one testcase. I've added one more testcase to it for now. It will take time to fully cover gun's functionality. Maybe we could merge this PR as is and write testcases as we go?

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -23,23 +23,32 @@
 %% Requests.
 -export([delete/2]).
 -export([delete/3]).
+-export([delete/4]).
 -export([get/2]).
 -export([get/3]).
+-export([get/4]).
 -export([head/2]).
 -export([head/3]).
+-export([head/4]).
 -export([options/2]).
 -export([options/3]).
+-export([options/4]).
 -export([patch/3]).
 -export([patch/4]).
+-export([patch/5]).
 -export([post/3]).
 -export([post/4]).
+-export([post/5]).
 -export([put/3]).
 -export([put/4]).
+-export([put/5]).
 -export([request/4]).
 -export([request/5]).
+-export([request/6]).
 
 %% Streaming data.
 -export([data/4]).
+-export([data/5]).
 
 %% Awaiting gun messages.
 -export([await/2]).
@@ -57,6 +66,7 @@
 
 %% Cancelling a stream.
 -export([cancel/2]).
+-export([cancel/3]).
 
 %% Websocket.
 -export([ws_upgrade/2]).
@@ -83,6 +93,9 @@
 
 -type opts() :: map().
 -export_type([opts/0]).
+
+-type req_opts() :: map().
+-export_type([req_opts/0]).
 
 -type http_opts() :: map().
 -export_type([http_opts/0]).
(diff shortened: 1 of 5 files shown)


essen:
Yes spdy_SUITE has tests for things that are not implemented. I'm considering just dropping SPDY entirely.

essen:
I'd say if there's a connection error, then all processes that have currently emitted a request are supposed to receive a gun_error, no?

manifest:
Good point! It would be match more expected behavior, I believe. I've updated the PR.

manifest:
It also might be useful to pass `ReplyTo :: pid()` to the request function. It allows to redirect response to any process we like. Do we care about backwards compatibility? Could we change arguments of `request/{4,5}` functions, or it would be better to introduce something like `request_to` functions?

```erlang
-spec request(pid(), ReplyTo :: pid(), iodata(), iodata(), headers()) -> reference().
-spec request(pid(), ReplyTo :: pid(), iodata(), iodata(), headers(), iodata()) -> reference().
```

essen:
I would be OK keeping things like this, and adding a function with +1 argument `Opts` that is a map that can have a value `reply_to` that takes a pid that defaults to `self()`. Then nothing breaks, and we have a way to extend things in the future.

essen:
I mean adding a function for each of request, get, post, put...

manifest:
So, instead of

```erlang
-spec post(pid(), iodata(), headers()) -> reference().
-spec post(pid(), iodata(), headers(), iodata()) -> reference().
```

we would have

```erlang
-spec post(pid(), iodata(), headers()) -> reference().
-spec post(pid(), iodata(), headers(), BodyOrOpts) -> reference() when BodyOrOpts :: map().
-spec post(pid(), iodata(), headers(), Body :: iodata(), Opts :: map()) -> reference().
```

Did I get it right?

essen:
Not `BodyOrOpts`, it's easier for users to just write `[]` than to reason about two different argument types. Just the third function.

manifest:
Ok, I got it. Just wonder, what do you think about this variant (to use map for all optional arguments):

```erlang
-type get_request_options() :: #{headers => list(), reply_to => pid()}.
-type put_request_options() :: #{headers => list(), body => iodata(), reply_to => pid()}.

-spec get(pid(), Path :: iodata()) -> reference().
-spec get(pid(), Path :: iodata(), Opts :: get_request_options()) -> reference().

-spec put(pid(), Path :: iodata()) -> reference().
-spec put(pid(), Path :: iodata(), Opts :: put_request_options()) -> reference().

essen:
These aren't options, they're requests with or without extra headers or bodies. It's part of the request, so it shouldn't be hidden like this.

manifest:
Ready. I've also added extended versions of `data/4` and `cancel/2`.

essen:
Let me try to close/reopen to see if it triggers the build.

essen:
Nope. Can I bother you to first: rebase to current master, and then: open a new PR? Thanks.

manifest:
Done.

PR #111, opened 2016-11-25
Sync rebar.config deps with Makefile

leoliu:


essen:
Actually don't, generate it from the Makefile instead. Add a line like this (with comment): https://github.com/ninenines/cowboy/blob/master/Makefile#L41 and then run `make` and commit what changed.

leoliu:
That's much nice. Thanks.

essen:
Looks good. Will merge soon. Thanks!

essen:
Merged, thanks!

PR #113, opened 2017-01-25
Ignore errors from Transport:setopts/2

mururu:
`Transport:setopts/2` can return an error after the socket has been closed. We should ignore it.


essen:
Merged, thanks!

PR #116, opened 2017-03-23
Feature/transform header name

### LINKED ISSUE #115: Ability to transform headers before they are sent
HTTP1 header names are supposed to be treated without regards to case. HTTP2 requires al header names to be lower case. Gun expects all header names to be lower case and depends on it in order to function correctly.

It would sound like the solution to the above is to simply lower-case all header names. But it is not that simple: many applications - and some servers - out there depend on the header name casing (see a similar discussion here: https://github.com/whatwg/fetch/issues/304).

The proposed solution (essen or IRC) is to give implementors the ability to transform header names via a callback before they are sent to a HTTP1 server.

strange:
This is the proposed solution to deal fix #115 

Please advice on how you want the type `transform_header_name_callback()` documented. I couldn't find a format that I was happy with.

### DIFF
--- a/src/gun_http.erl
+++ b/src/gun_http.erl
@@ -42,7 +42,8 @@
 	streams = [] :: [{reference() | websocket_info(), binary(), boolean()}],
 	in = head :: io(),
 	in_state :: {non_neg_integer(), non_neg_integer()},
-	out = head :: io()
+	out = head :: io(),
+	transform_header_name = undefined :: function() | undefined
 }).
 
 check_options(Opts) ->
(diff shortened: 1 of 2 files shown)


src/gun_http.erl
```diff
@@ -478,3 +487,9 @@ ws_handshake_end(Buffer, #http_state{owner=Owner, socket=Socket, transport=Trans
 			self() ! {OK, Socket, Buffer}
 	end,
 	gun_ws:init(Owner, Socket, Transport, Headers, Extensions, Protocols).
+
+apply_transform_header_name(#http_state{ transform_header_name = Fun }, Headers)
+  when is_function(Fun) ->
+	lists:keymap(fun(HeaderName) -> Fun(HeaderName) end, 1, Headers);
```
phanimahesh: Why not just `lists:keymap(Fun, 1, Headers)`?

essen:
I added a test, tweaked things a bit and merged. Thanks everyone!

PR #123, opened 2017-05-22
gun_ws checking protocols parameter always matches as error

shortishly:
[[is_binary(B), is_atom(M)] || {B, M} <- L] can't match to [true] resulting in always matching to the error clause. Simple change to change match to [true, true].

### DIFF
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -32,8 +32,12 @@
 	reply_to :: pid(),
 	%% Whether we finished sending data.
 	local = nofin :: cowboy_stream:fin(),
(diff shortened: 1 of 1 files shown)


essen:
Pretty sure this patch is wrong considering we do lists:flatten.

essen:
I retract that comment, what am I saying...

essen:
Your patch will also fail for more than one protocol. I think what's missing is a lists:usort call.

essen:
Just adding lists:usort to the existing code would be clearer.

essen:
Or lists:all as a one liner would be fine I guess. Not with that weird formatting.

PR #124, opened 2017-05-22
add missing call to lists:usort/1 in gun_ws:do_check_options/1

shortishly:
much better solution!


essen:
Thanks!

essen:
Merged, thanks!

PR #126, opened 2017-06-15
Adding in option to stop gun sending keepalive msgs between requests

sata:


essen:
Cheers I'm with customer but will try to get to it in a few days.

essen:
I've just pushed a more complete patch for this including allowing it for HTTP/2, documentation and a simple test. Thanks!

Issue #105, opened 2016-10-01
no function clause matching gun_http2:parse(<<>>,close)


leoliu:
Do `gun:open("fcm.googleapis.com", 443)` and wait a while to see SASL prints:

```
=CRASH REPORT==== 1-Oct-2016::11:15:02 ===
  crasher:
    initial call: gun:init/5
    pid: <0.3368.0>
    registered_name: []
    exception error: no function clause matching gun_http2:parse(<<>>,close) (/Users/leo/sources/erlang/test_gun/_build/default/lib/gun/src/gun_http2.erl, line 77) 
      in function  gun:loop/1 (/Users/leo/sources/erlang/test_gun/_build/default/lib/gun/src/gun.erl, line 584)
    ancestors: [gun_sup,<0.3142.0>]
    messages: []
    links: [<0.3143.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 1598
    stack_size: 27
    reductions: 4071
  neighbours:
```

leoliu:
This is due to `gun_http2:frame/2` called in `gun_http2:parse/2` may not return `##http2_state`.

essen:
Thanks! And sorry I did not see before, I am trying to solve a problem where I don't receive many emails from github.

manifest:
I've got the same issue after upgrading to latest vesion of Cowboy (2.0.0-pre.9).

manifest:
It hasn't reproduced locally, but always on Travis CI.
https://travis-ci.org/manifest/datastore/builds/243632587#L3204-L3213

edgurgel:
I get the same error after some time that the connection is up. 

```
{:DOWN, #Reference<0.0.2.22>, :process, #PID<0.204.0>,
  {:function_clause,
   [{:gun_http2, :parse, ["", :close], [file: 'src/gun_http2.erl', line: 106]},
    {:gun, :loop, 1, [file: 'src/gun.erl', line: 639]},
    {:gun, :proc_lib_hack, 5, [file: 'src/gun.erl', line: 528]},
    {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 247]}]}}]
```

I will try to send a PR later this week

edgurgel:
From what I can see this is the last frame sent before the error comes up:

```
{goaway,0,enhance_your_calm,<<"too_many_pings">>}
```

essen:
[...]

PR #131, opened 2017-08-27
Fix gun_http2:handle/2 when goaway is received

edgurgel:
It fixes https://github.com/ninenines/gun/issues/105

I'm not sure if this is correct approach and if I'm following the correct code style. 

I also added 1 eunit test but I'm also not sure what's the correct way to test this case.

### DIFF
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -103,18 +103,21 @@ init(Owner, Socket, Transport, Opts) ->
 handle(Data, State=#http2_state{buffer=Buffer}) ->
 	parse(<< Buffer/binary, Data/binary >>, State#http2_state{buffer= <<>>}).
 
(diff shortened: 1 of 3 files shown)


essen:
I think this is the right way to go. I'm not a fan of having eunit tests for this but we don't have a full fledged test server that can send whatever garbage we need yet so that'll do.

essen:
Merged, thanks!

edgurgel:
Yeah I couldn't find a way to test without adding an eunit test. 
Do you have any plans to have a test server? Would it be cowboy or some other implementation? 


Thanks for merging my PR! 🎉

essen:
I wouldn't say "plans" but it'll be necessary at some point. That's the main reason why Gun is lacking in the tests department. And it can't be Cowboy because Cowboy doesn't send garbage.

PR #132, opened 2017-09-19
Update links in README

KrzysiekJ:
Mailing list was archived on August 29, 2016: https://ninenines.eu/archives/extend/

http://ninenines.eu/support is now a redirect.


essen:
Merged, thanks!

PR #133, opened 2017-10-27
Modified a sentence

Chaitanya1416:


essen:
I think the original is correct.

Issue #134, opened 2017-12-07
Support for comms via Unix domain sockets


sanmiguel:
Since R19 we've had some support for opening unix domain sockets in erlang/otp. The underlying transport libraries support this, but it requires a little acrobatics to get everything working which is unfortunately precluded from working with `gun` as-is because of the `is_list(Host); is_atom(Host)` check in `gun:open/3`.

The minimum change would be to allow passing in a tuple `{local, SocketPath}` with a check for `is_list(SocketPath) ; is_binary(SocketPath)`. However that leaves more work for users to do:

- `Port` must be 0
- each request must manually pass the header `{<<"host">>, <<"localhost">>}`

I have [a branch](https://github.com/sanmiguel/gun/tree/http_unix_sockets) on which I've added a separate function, `gun:open_unix/2` that at least deals with the Host and Port. I'll submit a tentative PR shortly from it and we can take it from there.

essen:
Separate function sounds good.

PR #135, opened 2017-12-07
Support for HTTP over Unix domain sockets

sanmiguel:
As mentioned in #134 it's not that easy to use `gun` to make HTTP requests over a unix domain socket, but this branch aims to add that support. Please let me know if this approach is not one you would take and we can rework it.

I'd also like to find out how you would like to see this tested.

TODO:
 - [x] Confirm approach with @essen 
 - [x] Add explicit test cases for unix sockets
 - [x] Use a better `SocketPath` for test case

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -17,6 +17,7 @@
 %% Connection.
 -export([open/2]).
 -export([open/3]).
+-export([open_unix/2]).
 -export([info/1]).
 -export([close/1]).
 -export([shutdown/1]).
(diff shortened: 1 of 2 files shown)


essen:
Sounds good.

For the test I suppose creating a socket, connecting to it/sending a request and being able to receive something on the server side is enough. No need to parse or anything, just need to be able to connect and send stuff. Should be in `gun_SUITE`.

I'm guessing the test will need to be disabled in Windows.

For nitpicks I prefer `do_open` rather than `open_`, took me a little time to figure it out. `open_unix` should be a one liner call to `do_open` rather than creating variables.

sanmiguel:
@essen I've added a test case `unix_socket_connect/1`, but there's one detail I want to change: it uses a hard-coded path to the socket of `/tmp/gun.sock` which is obviously not ideal.

I'd like to use something from within CT's structure, but there's not already a `test/gun_SUITE_data` dir. Any thoughts on how you'd like it resolved?

I could do something like:

```
DataDir = config(data_dir, Config),
SocketPath = filename:join(DataDir, "gun.sock"),
filelib:ensure_dir(SocketPath),
_ = file:delete(SocketPath),
```

What do you reckon?

essen:
The `data_dir` is for files that are supposed to be committed to the repo, probably not what you want. I believe the one you want is `priv_dir`, which should be generated automatically per test run (the `log_private/` directory deep in the CT logs).

sanmiguel:
Ah, this is going to be fun... Was getting `{error, einval}` when using `priv_dir` instead...

From `man unix` on my mac:

```
UNIX-domain addresses are variable-length filesystem pathnames of at most 104 characters.
```

Current length of path to `priv_dir` including socket name: 163

How strongly would you object to using e.g. `test_server:temp_name("/tmp/gun")` to get the directory name?

Also, how do we go about disabling this test for Windows?

essen:
Oh then no problem just go with `/tmp`.

Just do something like https://github.com/ninenines/cowboy/blob/master/test/req_SUITE.erl#L163 except over `case os:type() of {win32, _} -> ...; _ -> ... end`.

sanmiguel:
Cool thanks, I'll get it done in the morning. :+1:

sanmiguel:
@essen this is now ready for you to look at! 👍

dcgibbons:
eagerly awaiting this one! 👍

essen:
Merged, thanks!

sanmiguel:
Thanks!

PR #148, opened 2018-04-15
add flow control

tony612:


essen:
I will try and merge it if it's good, but note that it'll be a temporary solution, I want to extract most of the functions in Cowboy and reuse them in Gun as a lot is shared between client/server.

essen:
Are you sure this is working? I can't seem to be able to send more than 65535 bytes.

essen:
I'm working on importing the relevant code from Cowboy for now. That'll have the benefit of having trailers support too.

essen:
Will push what I got last week in a few minutes. Only had to do minimal changes compared to what Cowboy does so that'll make it easy to split things off into a common module later on. Closing, thanks!

PR #155, opened 2018-05-28
Fix! handle domain socket host for http2 authority header

drowzy:
unix domain socket tuple matched as a IP and called `inet:ntoa`.


essen:
Half-patch, it's the same in `gun_http`.

essen:
Merged locally, push in a few. Thanks!

drowzy:
Super! Thanks

PR #156, opened 2018-06-11
Fix await_up link in function reference docs

rciorba:
I noticed the bad link text while reading the docs, so here's a tiny PR to fix it.

Thanks for the great libs, cowboy+gun are lovely to use!


essen:
Good catch, thanks!

essen:
Merged!

PR #159, opened 2018-07-03
feat: add frame function to handle headers

Draveness:


essen:
Not against the idea but please take note of this todo https://github.com/ninenines/gun/pull/159/files#diff-e27a5387bc74f9e3ff12cae2f3465124L191 and replace it with what you wrote instead.

essen:
Done locally with a test in a49200eef5681a21b47b9398804021b340486db8. It will be pushed later today. Thanks!

PR #164, opened 2018-08-31
Add links to Manual and User Guide to the README

filmor:


essen:
Thanks, I've added more links and instructions to my local copy and will push soon.

PR #181, opened 2018-12-17
Fix invalid stream reference in `gun_data` message for rejected websocket upgrade

pbober:
When a server rejects a websocket upgrade request and sends a response body, gun doesn't strip the stream reference (set to `{websocket, ...}`) in the `gun_data` message. This breaks matching the `StreamRef` value in these messages.

I added a unit test for this case (it seems that the Autobahn suite doesn't verify it).


essen:
Good find! Thanks. I'll review and merge over the holidays.

essen:
Merged locally, thanks!

pbober:
Thanks! Are you planning a new gun release any time soon?

PR #196, opened 2019-03-13
host header in Ws upgrade fixed

SergeTupchiy:
used already_implemented host_header/3 func, when adding host header in ws upgrade,
without it ws upgrade failed if Host was of  ip4_address() type: {0..255, 0..255, 0..255, 0..255}


essen:
Merged, thanks!

PR #197, opened 2019-04-19
WIP: Add support for connecting to TLS origins via TLS proxies

essen:
Still need to add ALPN support and to wait before trying to send
data on a proxied TLS connection that didn't complete its handshake.

Please experiment and provide feedback!


essen:
The feature should be more or less ready. Any review would be much appreciated!

essen:
Merged!

PR #199, opened 2019-04-28
Handle requests when in not_connected to wait for retrying

tony612:


essen:
Against the current code I'm thinking this should be enough:

``` erlang
index a548a5d..00c7cfd 100644
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -1061,6 +1062,9 @@ handle_common(cast, Any, _, #state{owner=Owner}) when element(2, Any) =/= Owner
    element(2, Any) ! {gun_error, self(), {notowner,
        "Operations are restricted to the owner of the connection."}},
    keep_state_and_data;
+%% We postpone all HTTP/Websocket operations until we are connected.
+handle_common(cast, _, StateName, _) when StateName =/= connected ->
+   {keep_state_and_data, postpone};
 handle_common(Type, Event, StateName, StateData) ->
    error_logger:error_msg("Unexpected event in state ~p of type ~p:~n~w~n~p~n",
        [StateName, Type, Event, StateData]),
```

Thoughts?

tony612:
@essen It's OK for me.

tony612:
@essen I added a test and changed it same as your changes. Feel free to use your own code if you already finish it.

essen:
I'm using your commit but rewrote the test to confirm the request actually goes through when the connection finally succeeds. Merged, thanks!

PR #200, opened 2019-04-28
Support a function as retry_timeout

tony612:
Support pass a function as retry_timeout so that we can support more flexible timeout like exponential backoff.

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -281,6 +281,8 @@ check_options([{retry, R}|Opts]) when is_integer(R), R >= 0 ->
 	check_options(Opts);
 check_options([{retry_timeout, T}|Opts]) when is_integer(T), T >= 0 ->
 	check_options(Opts);
+check_options([{retry_timeout, F}|Opts]) when is_function(F) ->
+	check_options(Opts);
 check_options([{supervise, B}|Opts]) when B =:= true; B =:= false ->
 	check_options(Opts);
 check_options([{tcp_opts, L}|Opts]) when is_list(L) ->
@@ -740,6 +742,7 @@ start_link(Owner, Host, Port, Opts) ->
 
 init({Owner, Host, Port, Opts}) ->
 	Retry = maps:get(retry, Opts, 5),
+	RetryTimeout = maps:get(retry_timeout, Opts, 5000),
 	OriginTransport = maps:get(transport, Opts, default_transport(Port)),
 	{OriginScheme, Transport} = case OriginTransport of
 		tcp -> {<<"http">>, gun_tcp};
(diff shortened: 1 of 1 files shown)


src/gun.erl
```diff
@@ -261,10 +261,14 @@ check_options([Opt = {protocols, L}|Opts]) when is_list(L) ->
 		_ ->
 			{error, {options, Opt}}
 	end;
+check_options([{retry, infinity}|Opts]) ->
```
essen: That's unrelated and I'm not sure why you added this? A sufficiently large value would be good enough. And even if you want to *always* reconnect there's a point where Gun should give up because something is really wrong.

essen:
Probably worth returning both timeout and number of retries left instead, maybe wrapped in a map.

essen:
I also would rather pass the options to the fun as well, and use a separate option for the function.

essen:
Something like this maybe:

``` erlang
index bb275f5..cf6e2ef 100644
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -116,6 +116,7 @@
    http2_opts => http2_opts(),
    protocols => [http | http2],
    retry => non_neg_integer(),
+   retry_fun => fun((non_neg_integer(), opts()) -> #{retries => non_neg_integer(), timeout => pos_integer()}),
    retry_timeout => pos_integer(),
    supervise => boolean(),
    tcp_opts => [gen_tcp:connect_option()],
@@ -770,15 +771,25 @@ default_transport(_) -> tcp.
 %% @todo This is where we would implement the backoff mechanism presumably.
 not_connected(_, {retries, 0, Reason}, State) ->
    {stop, {shutdown, Reason}, State};
-not_connected(_, {retries, Retries, _}, State=#state{opts=Opts}) ->
-   Timeout = maps:get(retry_timeout, Opts, 5000),
+not_connected(_, {retries, Retries0, _}, State=#state{opts=Opts}) ->
+   Fun = maps:get(retry_fun, Opts, fun default_retry_fun/2),
+   #{
+       timeout := Timeout,
+       retries := Retries
+   } = Fun(Retries0, Opts),
    {next_state, domain_lookup, State,
-       {state_timeout, Timeout, {retries, Retries - 1, not_connected}}};
+       {state_timeout, Timeout, {retries, Retries, not_connected}}};
 not_connected({call, From}, {stream_info, _}, _) ->
    {keep_state_and_data, {reply, From, {error, not_connected}}};
 not_connected(Type, Event, State) ->
    handle_common(Type, Event, ?FUNCTION_NAME, State).
 
+default_retry_fun(Retries, Opts) ->
+   #{
+       timeout => maps:get(retry_timeout, Opts, 5000),
+       retries => Retries - 1
+   }.
+
```

tony612:
@essen I solved the conflicts and updated the code as you said. Tests will be added later.

Updates: Just noticed your example code. I'll think about it later.

essen:
If my snippet is good for you I'll continue on my own with this.

essen:
I will also make it so the state_timeout isn't applied if retries is returned as 0. It'll just shut down immediately.

essen:
Hm no nevermind.

tony612:
@essen How about adding a field for current retry times(or total retries)? I'm OK with your snippet.

essen:
The fun receives the current number of retries and the options (including retry/retry_timeout) and returns the next number of retries and the timeout to use. I think this includes everything you need?

essen:
It also allows you to have infinite retries, just never return 0.

tony612:
@essen 
> I think this includes everything you need?
Yes. Thank you!

essen:
Pushed. Thanks!

PR #201, opened 2019-05-21
Replace non-existant types

bblaszkow06:
There is no such type as `ssl:connect_option` I've replaced it with `ssl:tls_client_option`. You can check [this](http://erlang.org/doc/man/ssl.html#type-server_option) for reference.

This fixes `unknown type` warning from dialyzer.


essen:
Thanks I've fixed it in a few more places and in the documentation. It'll be pushed later today.

PR #219, opened 2020-03-02
Add support for configurable cookie stores

essen:
Implementing much of https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#ua-requirements

Also need https://github.com/ninenines/cowlib/pull/96

Only the documentation is left. Perhaps some additional tests but this is becoming pretty good already.

/cc @petrohi


essen:
I've just pushed a WIP commit where Gun is actually using the cookie store, and some of the WPT tests pass (all http-state ones pass for HTTP, most pass for HTTPS, none pass for HTTP/2).

essen:
As far as the code goes, this is more or less complete. The documentation remains to be written.

PR #238, opened 2020-08-11
fix rebar3 dependency specification

galdor:
Rebar3 emits a warning for "raw" git dependency versions, and expects a
tuple indicating what the version actually is (a branch, tag or ref).

This warning propagates to any software directly or indirectly using
Gun.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
 {deps, [
-{cowlib,".*",{git,"https://github.com/ninenines/cowlib","2.9.0"}}
+{cowlib,".*",{git,"https://github.com/ninenines/cowlib",{tag,"2.9.0"}}}
 ]}.
 {erl_opts, [debug_info,warn_export_vars,warn_shadow_vars,warn_obsolete_guard]}.


essen:
This file is generated: https://github.com/ninenines/gun/blob/master/Makefile#L49

It's also like this to be compatible with Rebar 2. I don't know if assuming everyone uses Rebar 3 is a good idea. Some people definitely still use Rebar 2. I don't know if Rebar 2 supports this tuple. If it does the fix should be in Erlang.mk.

galdor:
Rebar3 is now part of the erlang GitHub project. Rebar2 is officially deprecated, there has not been any patch for two years and a half, and the repository has been archived. Of course some people will never upgrade anything, I'm not sure to see the point in supporting them.

In any case, Rebar2 [supports](https://github.com/rebar/rebar/blob/master/src/rebar_deps.erl#L608) this notation.

Anyway I just open a [new PR](https://github.com/ninenines/gun/pull/239) which updates erlang.mk.

essen:
OK then, please just send the patch to the right location. You can confirm it works by running `make check c=core-compat` as well.

PR #239, opened 2020-08-11
fix rebar3 dependency specification

galdor:
Rebar3 emits a warning for "raw" git dependency versions, and expects a
tuple indicating what the version actually is (a branch, tag or ref).

This warning propagates to any software directly or indirectly using
Gun.


essen:
I meant the change should be against the Erlang.mk project, here specifically: https://github.com/ninenines/erlang.mk/blob/master/core/compat.mk

There's a pending update of Erlang.mk for Gun and other projects (see other PRs) so the erlang.mk file will be updated at the same time.

PR #241, opened 2020-08-18
Enabling h2specd_SUITE.erl again

bjosv:
The main source file for the test tool h2specd was moved a time ago, so this PR make sure we use correct path when building it.
The h2spec project also uses `go modules` which now is used when building.

The change has been verified (build and passing of h2specd_SUITE) in a Github Action setup that uses a test matrix, consisting of all major versions of Go from v1.11 (2018) to latest 1.15.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ $(H2SPECD):
 	$(gen_verbose) mkdir -p $(GOPATH)/src/github.com/summerwind
 	-$(verbose) git clone --depth 1 https://github.com/summerwind/h2spec $(dir $(H2SPECD))
 	-$(verbose) $(MAKE) -C $(dir $(H2SPECD)) build MAKEFLAGS=
-	-$(verbose) go build -o $(H2SPECD) $(dir $(H2SPECD))/cmd/h2spec/h2specd.go
+	-$(verbose) cd $(dir $(H2SPECD)) && go build cmd/h2specd/h2specd.go
 
 # Public suffix module generator.
 # https://publicsuffix.org/list/
--- a/test/h2specd_SUITE.erl
+++ b/test/h2specd_SUITE.erl
@@ -97,7 +97,8 @@ run_tests([Port|Tail]) ->
 	try
 		{ok, Conn} = gun:open("127.0.0.1", Port, #{
 			protocols => [http2],
-			retry => 0
+			retry => 0,
+			tcp_opts => [{nodelay, true}]
 		}),
 		MRef = monitor(process, Conn),
 		{ok, http2} = gun:await_up(Conn),


bjosv:
Well, this didn't go as expected..
Running local or using Github Actions seems to give different results for h2specd_SUITE.erl , I'll have to investigate...

**Base/Master**
- Arch Linux - fails to remove gopath
- Alpine        - makefile fails
- CentOS     - fails to remove gopath
- Debian      - makefile fails
- macOS      - Go not available
- Ubuntu         - makefile fails
- Windows 10 - Go not available

**Deviating buildkite results in this PR**
- Alpine     -  h2specd_SUITE.erl fails
- Debian     - h2specd_SUITE.erl fails

essen:
Right there's some BS about Go fetching things and putting them read-only or something. I may or may not have fixed that issue in Cowboy.

bjosv:
Yeah, the handling of it in Go seems a bit weird too.. the reason for having it read-only is to make sure tests wont write logs amongst source files..

The goto recommendation was to use `go clean -modcache` to remove files, but in a late version of go (1.14, Feb 2020) they finally added the `-modcacherw` flag that makes sure the fetched files are read-writable.
Adding this flag is probably the correct way of making sure files can be removed easily, but will put tight version restrictions on the build of this test. Any recommendations?

I also checked Cowboy, and if I understand the build-logs correctly the problem seems to be avoided by changing the Buildkite pipeline config, possibly via BUILDKITE_CLEAN_CHECKOUT and BUILDKITE_GIT_CLEAN_FLAGS.
I guess I cant change this my self.

essen:
Went to check. In Cowboy I simply set `GOFLAGS=-modcacherw` as environment variable. So I guess it's OK to do the same for Gun. And in fact, it was already set for Gun, but not for PR builds (and not for Cowboy PR builds).

Try again? Or do we need to delete the files first?

bjosv:
I think it will fail due to the initial git clone would fail again.
In cowboy this step differs, it does a git remote and git clean first...so that why I thought the pipeline differs regarding BUILDKITE_CLEAN_CHECKOUT or something similar

essen:
Maybe you're comparing normal Cowboy builds and Gun PR builds. PR instructions are different. I don't set anything special.

I've removed the files, you can try again.

bjosv:
Ah, I see. Found the environment tab in Buildkit now..

Looks like files are still blocking for`alpine` and `debian`, those targets seems to have older Go version anyway (unknown flag) but should get green after removing the old files.
`Arch Linux` is the only target that can build and run the testsuite, but unfortunately some tests in h2specd_SUITE are failing..are there any known issues, or something special with the machine? Like 32bit.. I keep digging in the testsuite..

essen:
No known issues in that it used to work but nothing's forever. I'll clean up alpine and debian.

essen:
I have updated the Ubuntu environment to 20.04 which has Go 1.13. Please retry.

bjosv:
Super, kicked of a retry.

essen:
Since upstream is taking such a long time perhaps we could do this PR in 2 parts: first part to fix the running issue itself, and second part about the intermittent issue. I could merge the first part immediately.

bjosv:
Sounds like a plan, I changed so that we now run the h2specd from its source,
and hopefully the intermittent issue is not that frequent in CI tests.

essen:
Merged, thanks!

PR #242, opened 2020-09-03
Include content_handlers in opts

tomas-abrahamsson:
This fixes a dialyzer issue when using the gun and setting the `content_handlers` option.

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -184,6 +184,7 @@
 
 -type http_opts() :: #{
 	closing_timeout => timeout(),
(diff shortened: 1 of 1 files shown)


tomas-abrahamsson:
Oops, just force-pushed; fixed a `gun_content_handler` module name.

essen:
Hey, I would like to document content handlers at the same time, but there's still an unknown that should be resolved first (trailers should probably be passed to content handlers).

essen:
Heh going to merge it now anyway. Thanks!

Issue #224, opened 2020-04-05
Set a default send_timeout value


essen:
Sometimes the client is sending much faster than the server is able to process, and the client gets stuck sending data. Then the server may also get stuck when sending data in responses that the client will not process, for example when HTTP/1.1 pipelining was used. Gun should not lock up in these cases, it should just give up.

The Transport:send return value may also need to be checked in order to speed up dropping the connection.

essen:
Done.

PR #243, opened 2020-09-16
Handle send errors

### LINKED ISSUE #227: The GUN instance still exist after the socket was closed by OS(in the case of IP path failure).
We open a GUN instance successfully with the following TCP options {send_timeout, 30000} and {send_timeout_close, true}.
The options makes if the TCP stack can’t receive the sending packet and return a timeout, the socket will be closed.
Once the IP path has problem on the TCP connections, and the upper application sends the http request to the GUN instance continuously, the socket will closed finally.
It may be caused by the TCP sending window/buffer full.  Then the OS close the socket will and send the TCP RST to peer(although it is impossible received by peer host).
However, the GUN instance is insensitive with this event. In the function gun_http2:request(), the returned value was not checked for gen_tcp:send().
The return value should be {error,timeout} first and {error,enotconn} after the socket is closed. I think GUN instance should exit for reason {error,enotconn} anyway.

bjosv:
[To get faster reaction times on failing socket-send of echos/requests
we now handle the response from the send call where applicable.

To avoid a larger change in the statemachine a send error triggers an
event and is handled like a remote socket error/close.

Fixes #227 and touches #224

This PR is a proposal and should be seen as a base for discussing how to solve it the best way,
and ideas/comments are welcome. 
(http and ws should have the same final solution, and tests can be cleaned up a bit aswell..)


essen:
I suggest waiting for my HTTP/2 CONNECT (http2-connect branch) work to be complete because the socket will not always be a real socket and I suspect that this solution will not work in that case. I also dislike using a message, I'd rather return a command, though let's first see what the implications would be for proxies (TLS proxies may close the connection at the TLS level and this needs to be handled in a different way than when the real socket closes).

zuiderkwast:
Hi! I'm collaborating with @bjosv on this. Do you want us to rebase this PR on the http2-connect branch?

This PR just addresses the fact that the return value from `Transport:send/2` is ignored. I can't find anything related to that in the http2-connect branch. If the socket is not a real socket, Transport:send/2 can still return `ok | {error, Reason}` (since it's just what gen_tcp:send/2 and ssl:send/2 return, right?).

The message passing to `self()` is perhaps a bit ugly, but since most other errors from the transport come in the form of message passing, passing the error returned by send as a message to self makes gun handle it just like any other error, which is quite nice. Of course it would be better to abort sending a request once sending the headers fails, etc.

Can you clarify "I'd rather return a command"? Do you mean gun_http2:request, gun_http2:maybe_send_data, etc. should return something to trigger a new state in the gun state machine?

essen:
Do not rebase against http2-connect, it's still undergoing significant changes.

For HTTP/2 CONNECT, yes the fake sockets still return errors (potentially, anyway), however the error has to be routed to the correct proxy layer (so if you have, for example, the socket at the end of the HTTP/2 CONNECT -> HTTP/2 CONNECT -> HTTP/1.1 returning an error, it must be handled by the middle layer, not by the innermost layer.

> Can you clarify "I'd rather return a command"? Do you mean gun_http2:request, gun_http2:maybe_send_data, etc. should return something to trigger a new state in the gun state machine?

Yes. There is already an `{error, _}` command for example. But because of the large changes coming up it's best to wait a little before doing the work.

essen:
Hello, I see you probably noticed that I merged the http2-connect branch. :-)

Note that I don't think sending ourselves a message is going to work well due to tunnels. I'm going to see if making all callbacks follow the `{Commands, EvHandlerState}` format would work (necessary for proper event handling as well), in which case an `error` command should be returned instead.

essen:
I think work on this can be resumed. Like I've said earlier an error command should be returned. If it's in a callback that currently returns `{State, EvHandlerState}` it can "easily" be converted to `{Commands, EvHandlerState}`. I think the `update_window` callback would be in that case.

zuiderkwast:
This one would be good to have in 2.0. I'll give it a try.

zuiderkwast:
I've done a large rewrite now to propagate the errors as commands.

A command is returned instead of a state by these Protocol functions:

* headers/12
* request/13
* keepalive/3
* data/7
* connect/9
* ws_upgrade/11
* cancel/5

In gun_http2, several functions may return a state or an error instead of only a state, e.g.

* frame/4 (called by parse/5 in the same module)
* maybe_ack_or_notify/2 (called by frame/4)
* data_frame/7 -> data_frame1/7 (called by frame/4)
* update_window/1 (called by frame/4, data_frame1/7)
* update_window/2 (called by update_flow/4, data_frame1/7)
* reset_stream/3
* push_promise_frame/7
* goaway/2
* maybe_send_data/6 -> send_data/4 -> send_data/6 -> send_data_frame/4
* tunnel_commands/5

Also gun_http and other modules are updated. I think "http2" can be removed from the PR name since send errors are handled for all protocols.

essen:
Sounds good. But it's too big for me to look at today. If there's minor issues I will fix them during merge.

essen:
There's more than minor issues. The todos need to be resolved. But perhaps you'd prefer for me to do the implementation for `gun:cancel` of tunnel streams? Because then you could use the same mechanism to abort tunnel streams when a send error occurs (only in that case it would be aborted because of a socket error).

zuiderkwast:
> perhaps you'd prefer for me to do the implementation for `gun:cancel` of tunnel streams? Because then you could use the same mechanism to abort tunnel streams when a send error occurs

Yes, that would be good. I must confess I'm a bit lost in the tunnels. I'd be happy if you can give me some pointers.

essen:
There's a page in the user guide about the internals of the implementation. Please read it and tell me if this helps at all. I can improve while it's still clear in my mind.

zuiderkwast:
Do you mean "TLS over TLS"? I have seen it but I don't fully understand how things should be propagated.

essen:
Well the question is less about propagating and more about handling send errors also for the fake sockets, and what to do when that happens. Probably only stop the relevant stream for HTTP/2. But the entire connection for HTTP/1.1. Hence why I mention `gun:cancel` for tunnel streams, because then we know how to stop only tunnel streams.

bjosv:
Hi @essen , any more thoughts about the `gun:cancel` for tunnel streams?

Also, do you have any thoughts about incorporating this in steps? I guess we could incorporate the error propagation support first, then add send-error support for each protocol in steps. Or do you see a scenario that hinders this?
Current behavior is that the socket finally gets closed and this will close all tunnels anyway, similar to the PR.

src/gun.erl
```diff
@@ -1244,32 +1252,31 @@ connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow
 		State=#state{origin_host=Host, origin_port=Port,
 			protocol=Protocol, protocol_state=ProtoState, cookie_store=CookieStore0,
 			event_handler=EvHandler, event_handler_state=EvHandlerState0}) ->
-	{ProtoState2, CookieStore, EvHandlerState} = Protocol:headers(ProtoState,
+	{Commands, CookieStore, EvHandlerState} = Protocol:headers(ProtoState,
```
essen, line 1255: This.

src/gun.erl
```diff
@@ -1244,32 +1252,31 @@ connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow
...
 		dereference_stream_ref(StreamRef, State), ReplyTo,
 		Method, Host, Port, Path, Headers,
 		InitialFlow, CookieStore0, EvHandler, EvHandlerState0),
-	{keep_state, State#state{protocol_state=ProtoState2, cookie_store=CookieStore,
-		event_handler_state=EvHandlerState}};
+	commands(Commands, State#state{cookie_store=CookieStore,
+		event_handler_state=EvHandlerState});
```
essen, line 1260: This.

src/gun.erl
```diff
@@ -1244,32 +1252,31 @@ connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow
...
 connected(cast, {request, ReplyTo, StreamRef, Method, Path, Headers, Body, InitialFlow},
 		State=#state{origin_host=Host, origin_port=Port,
 			protocol=Protocol, protocol_state=ProtoState, cookie_store=CookieStore0,
 			event_handler=EvHandler, event_handler_state=EvHandlerState0}) ->
-	{ProtoState2, CookieStore, EvHandlerState} = Protocol:request(ProtoState,
+	{Commands, CookieStore, EvHandlerState} = Protocol:request(ProtoState,
```
essen, line 1265: This.

src/gun.erl
```diff
@@ -1244,32 +1252,31 @@ connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow
...
 		dereference_stream_ref(StreamRef, State), ReplyTo,
 		Method, Host, Port, Path, Headers, Body,
 		InitialFlow, CookieStore0, EvHandler, EvHandlerState0),
-	{keep_state, State#state{protocol_state=ProtoState2, cookie_store=CookieStore,
-		event_handler_state=EvHandlerState}};
+	commands(Commands, State#state{cookie_store=CookieStore,
+		event_handler_state=EvHandlerState});
```
essen, line 1270: This.

src/gun.erl
```diff
@@ -1244,32 +1252,31 @@ connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow
...
 connected(cast, {connect, ReplyTo, StreamRef, Destination, Headers, InitialFlow},
 		State=#state{origin_host=Host, origin_port=Port,
 			protocol=Protocol, protocol_state=ProtoState,
 			event_handler=EvHandler, event_handler_state=EvHandlerState0}) ->
-	{ProtoState2, EvHandlerState} = Protocol:connect(ProtoState,
+	{Commands, EvHandlerState} = Protocol:connect(ProtoState,
```
essen, line 1275: This.

src/gun.erl
```diff
@@ -1244,32 +1252,31 @@ connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow
...
 		dereference_stream_ref(StreamRef, State), ReplyTo,
 		Destination, #{host => Host, port => Port},
 		Headers, InitialFlow, EvHandler, EvHandlerState0),
-	{keep_state, State#state{protocol_state=ProtoState2,
-		event_handler_state=EvHandlerState}};
+	commands(Commands, State#state{event_handler_state=EvHandlerState});
```
essen, line 1279: This.

src/gun.erl
```diff
@@ -1284,11 +1291,11 @@ connected(cast, {ws_upgrade, ReplyTo, StreamRef, Path, Headers, WsOpts},
 		opts => WsOpts
 	}, EvHandlerState0),
 	%% @todo Can fail if HTTP/1.0.
-	{ProtoState2, CookieStore, EvHandlerState} = Protocol:ws_upgrade(ProtoState,
+	{Commands, CookieStore, EvHandlerState} = Protocol:ws_upgrade(ProtoState,
```
essen, line 1294: This.

src/gun.erl
```diff
@@ -1284,11 +1291,11 @@ connected(cast, {ws_upgrade, ReplyTo, StreamRef, Path, Headers, WsOpts},
...
 		dereference_stream_ref(StreamRef, State), ReplyTo,
 		Host, Port, Path, Headers, WsOpts, CookieStore0, EvHandler, EvHandlerState1),
-	{keep_state, State#state{protocol_state=ProtoState2, cookie_store=CookieStore,
-		event_handler_state=EvHandlerState}};
+	commands(Commands, State#state{cookie_store=CookieStore,
+		event_handler_state=EvHandlerState});
```
essen, line 1298: This.

src/gun.erl
```diff
@@ -1365,10 +1372,10 @@ closing(Type, Event, State) ->
 handle_common_connected(cast, {data, ReplyTo, StreamRef, IsFin, Data}, _,
 		State=#state{protocol=Protocol, protocol_state=ProtoState,
 			event_handler=EvHandler, event_handler_state=EvHandlerState0}) ->
-	{ProtoState2, EvHandlerState} = Protocol:data(ProtoState,
+	{Commands, EvHandlerState} = Protocol:data(ProtoState,
```
essen, line 1375: This.

src/gun.erl
```diff
@@ -1365,10 +1372,10 @@ closing(Type, Event, State) ->
...
 		dereference_stream_ref(StreamRef, State),
 		ReplyTo, IsFin, Data, EvHandler, EvHandlerState0),
-	{keep_state, State#state{protocol_state=ProtoState2, event_handler_state=EvHandlerState}};
+	commands(Commands, State#state{event_handler_state=EvHandlerState});
```
essen, line 1378: This.

src/gun.erl
```diff
@@ -1412,21 +1419,21 @@ handle_common_connected_no_input(info, {handle_continue, StreamRef, Msg}, _,
...
 handle_common_connected_no_input(info, keepalive, _,
-		State=#state{protocol=Protocol, protocol_state=ProtoState0,
+		State0=#state{protocol=Protocol, protocol_state=ProtoState0,
 		event_handler=EvHandler, event_handler_state=EvHandlerState0}) ->
-	{ProtoState, EvHandlerState} = Protocol:keepalive(ProtoState0, EvHandler, EvHandlerState0),
-	{keep_state, keepalive_timeout(State#state{
-		protocol_state=ProtoState, event_handler_state=EvHandlerState})};
+	{Commands, EvHandlerState} = Protocol:keepalive(ProtoState0, EvHandler, EvHandlerState0),
+	commands(Commands, keepalive_timeout(State0#state{
+		event_handler_state=EvHandlerState}));
```
essen, line 1426: This.

src/gun.erl
```diff
@@ -1412,21 +1419,21 @@ handle_common_connected_no_input(info, {handle_continue, StreamRef, Msg}, _,
...
 	Commands = Protocol:update_flow(ProtoState, ReplyTo, StreamRef, Flow),
 	maybe_active(commands(Commands, State0));
 handle_common_connected_no_input(cast, {cancel, ReplyTo, StreamRef}, _,
 		State=#state{protocol=Protocol, protocol_state=ProtoState,
 		event_handler=EvHandler, event_handler_state=EvHandlerState0}) ->
-	{ProtoState2, EvHandlerState} = Protocol:cancel(ProtoState,
+	{Commands, EvHandlerState} = Protocol:cancel(ProtoState,
 		dereference_stream_ref(StreamRef, State), ReplyTo, EvHandler, EvHandlerState0),
-	{keep_state, State#state{protocol_state=ProtoState2, event_handler_state=EvHandlerState}};
+	commands(Commands, State#state{event_handler_state=EvHandlerState});
```
essen, line 1436: This.

src/gun_http.erl
```diff
@@ -546,45 +546,47 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
 	close_streams(State, Tail, Reason).
 
 %% We don't send a keep-alive when a CONNECT request was initiated.
-keepalive(State=#http_state{streams=[#stream{ref={connect, _, _}}]}, _, EvHandlerState) ->
-	{State, EvHandlerState};
+keepalive(#http_state{streams=[#stream{ref={connect, _, _}}]}, _, EvHandlerState) ->
+	{[], EvHandlerState};
```
essen, line 550: This.

src/gun_http.erl
```diff
@@ -546,45 +546,47 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
...
-	{State, EvHandlerState};
-keepalive(State, _, EvHandlerState) ->
-	{State, EvHandlerState}.
+keepalive(#http_state{socket=Socket, transport=Transport, out=head}, _, EvHandlerState) ->
+	case Transport:send(Socket, <<"\r\n">>) of
+		ok -> {[], EvHandlerState};
+		Error={error, _} -> {Error, EvHandlerState}
+	end;
+keepalive(_State, _, EvHandlerState) ->
+	{[], EvHandlerState}.
```
essen, line 558: This last clause as well as returning `[]` in the clause before.

src/gun_http.erl
```diff
@@ -546,45 +546,47 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
...
+	{Authority, Conn, Out, CookieStore, EvHandlerState} = send_request(State0,
 		StreamRef, ReplyTo, Method, Host, Port, Path, Headers, undefined,
 		CookieStore0, EvHandler, EvHandlerState0, ?FUNCTION_NAME),
 	InitialFlow = initial_flow(InitialFlow0, Opts),
-	{new_stream(State#http_state{connection=Conn, out=Out}, StreamRef, ReplyTo,
-		Method, Authority, Path, InitialFlow),
-		CookieStore, EvHandlerState}.
+	State = new_stream(State0#http_state{connection=Conn, out=Out}, StreamRef,
+		ReplyTo, Method, Authority, Path, InitialFlow),
+	{{state, State}, CookieStore, EvHandlerState}.
```
essen, line 574: These.

src/gun_http.erl
```diff
@@ -546,45 +546,47 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
...
 
 request(State, StreamRef, ReplyTo, _, _, _, _, _, _, _, CookieStore, _, EvHandlerState)
 		when is_list(StreamRef) ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef),
 		{badstate, "The stream is not a tunnel."}},
-	{State, CookieStore, EvHandlerState};
+	{[], CookieStore, EvHandlerState};
```
essen, line 580: This.

src/gun_http.erl
```diff
@@ -546,45 +546,47 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
...
 		Method, Host, Port, Path, Headers, Body,
 		InitialFlow0, CookieStore0, EvHandler, EvHandlerState0) ->
 	{Authority, Conn, Out, CookieStore, EvHandlerState} = send_request(State,
 		StreamRef, ReplyTo, Method, Host, Port, Path, Headers, Body,
 		CookieStore0, EvHandler, EvHandlerState0, ?FUNCTION_NAME),
 	InitialFlow = initial_flow(InitialFlow0, Opts),
-	{new_stream(State#http_state{connection=Conn, out=Out}, StreamRef, ReplyTo,
-		Method, Authority, Path, InitialFlow),
+	{{state, new_stream(State#http_state{connection=Conn, out=Out}, StreamRef,
+		ReplyTo, Method, Authority, Path, InitialFlow)},
```
essen, line 589: And this.

src/gun_http.erl
```diff
@@ -689,41 +692,52 @@ data(State=#http_state{socket=Socket, transport=Transport, version=Version,
...
-					{State#http_state{out=head}, EvHandlerState};
+					case Transport:send(Socket, DataToSend) of
+						ok ->
+							RequestEndEvent = #{
+								stream_ref => stream_ref(State, StreamRef),
+								reply_to => ReplyTo
+							},
+							EvHandlerState = EvHandler:request_end(RequestEndEvent,
+								EvHandlerState0),
+							{{state, State#http_state{out=head}}, EvHandlerState};
```
essen, line 712: Only the bit changing the return value to `{state, State...}`.

src/gun_http.erl
```diff
@@ -689,41 +692,52 @@ data(State=#http_state{socket=Socket, transport=Transport, version=Version,
...
+						Error={error, _} ->
+							{Error, EvHandlerState0}
+					end;
 				body_chunked when Version =:= 'HTTP/1.1' ->
-					Transport:send(Socket, cow_http_te:chunk(Data)),
-					{State, EvHandlerState0};
```
essen, line 709: Only the bit returning `[]`.

src/gun_http.erl
```diff
@@ -689,41 +692,52 @@ data(State=#http_state{socket=Socket, transport=Transport, version=Version,
...
 								stream_ref => stream_ref(State, StreamRef),
 								reply_to => ReplyTo
 							},
 							EvHandlerState = EvHandler:request_end(RequestEndEvent, EvHandlerState0),
-							{State#http_state{out=head}, EvHandlerState};
-						Length2 > 0, IsFin =:= nofin ->
-							{State#http_state{out={body, Length2}}, EvHandlerState0}
+							{{state, State#http_state{out=head}}, EvHandlerState};
+						ok when Length2 > 0, IsFin =:= nofin ->
+							{{state, State#http_state{out={body, Length2}}}, EvHandlerState0};
```
essen, line 732: Only the bits about `{state...}`.

src/gun_http.erl
```diff
@@ -689,41 +692,52 @@ data(State=#http_state{socket=Socket, transport=Transport, version=Version,
...
+						Error={error, _} ->
+							{Error, EvHandlerState0}
 					end;
 				body_chunked -> %% HTTP/1.0
-					Transport:send(Socket, Data),
-					{State, EvHandlerState0}
+					case Transport:send(Socket, Data) of
+						ok -> {[], EvHandlerState0};
```
essen, line 738: Only the bit returning `[]`.

src/gun_http.erl
```diff
@@ -733,12 +747,12 @@ connect(State, StreamRef, ReplyTo, _, _, _, _, _, EvHandlerState)
 		when is_list(StreamRef) ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef),
 		{badstate, "The stream is not a tunnel."}},
-	{State, EvHandlerState};
+	{[], EvHandlerState};
```
essen, line 750: This.

src/gun_http.erl
```diff
@@ -733,12 +747,12 @@ connect(State, StreamRef, ReplyTo, _, _, _, _, _, EvHandlerState)
...
 connect(State=#http_state{streams=Streams}, StreamRef, ReplyTo, _, _, _, _, _, EvHandlerState)
 		when Streams =/= [] ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef), {badstate,
 		"CONNECT can only be used with HTTP/1.1 when no other streams are active."}},
-	{State, EvHandlerState};
+	{[], EvHandlerState};
```
essen, line 755: This.

essen (commented):
Let's do this in steps. I have pointed out in review comments some of the bits that I think should be merged as a first step. The list is not exhaustive. Could you open a PR with these types of changes? They relate to returning commands instead of state. Any other change should be left for step 2 (details TBD).

zuiderkwast:
> Let's do this in steps. I have pointed out in review comments some of the bits that I think should be merged as a first step. The list is not exhaustive. Could you open a PR with these types of changes? They relate to returning commands instead of state.

Done. Please have a look here: #285.

zuiderkwast:
Rebased.

essen:
Thanks. Will go over this probably on Thursday.

src/gun.erl
```diff
@@ -1233,15 +1233,23 @@ connected(internal, {connected, Socket, NewProtocol},
 		State0=#state{owner=Owner, opts=Opts, transport=Transport}) ->
 	{Protocol, ProtoOpts} = gun_protocols:handler_and_opts(NewProtocol, Opts),
 	{StateName, ProtoState} = Protocol:init(Owner, Socket, Transport, ProtoOpts),
+	State1 = State0#state{socket=Socket, protocol=Protocol, protocol_state=ProtoState},
```
essen, line 1236: I'm not happy with this part, `StateName` or `{error,_}` return value. It's probably made my opinion worse that it's not handled in many parts of the code, though. I wonder if there's a better way to handle cases where we have to send something immediately. I haven't had time to think much about it today unfortunately.

zuiderkwast: Is it specifically the return value of `Protocol:init/4` you're concerned with?

I don't know how to better represent it, bit I think init needs to be able to return an error. The related the todos can probably be handled by disconnecting as we do in `commands/2` ([gun.erl:1625)[https://github.com/ninenines/gun/blob/2.0.0-rc.2/src/gun.erl#L1625]) WDYT?

```Erlang
commands([Error={error, _}|_], State) ->
	disconnect(State, Error);
```

essen: I need to read more of the Gun code to answer, maybe reorganize a few things. Might take a while as I got a RabbitMQ bug to investigate.

essen: OK first I note that the code in this and the switch_protocol command are very similar. So we should work toward making them share the same code. The difference is that one sends a message, the other triggers an event. The shared code should start at Proto:init until the end of the clause. The message/event can be sent before returning next_state instead of where it is sent now (it could be a function that receives the updated State and returns a final State variable).

Perhaps this could be done as an intermediate state (or two) with a forced event that contains the function or similar.

The init functions should either return `{ok, StateName, ProtoState}` or `{error, Reason}`. Then we can match on THAT rather than on `StateName` and make the code more obvious. And later we could refactor to make use of `maybe` (especially for calls in protocols). The rest of the changes should be updated to have ok/error tuples (I see some `State | {error, _}` in `gun_http2` for example).

For the Proto:init/4 and init/6 we need to figure out how they should be handled. I do not think this should be merged until they have been figured out. I can help figure out what to do for the tunnels (some must bubble up to the real socket disconnecting, others just close a stream with optionally a fake socket).

I believe the first part (refactor) can be done separately. I think it would help also to do this on a per-protocol basis, starting with the simplest. A first easy PR would be this small refactor + easy protocols (raw, Websocket, socks, whatever takes over the entire connection). I think `gun_http` shouldn't be too difficult on top of that. `gun_http2` and `gun_tunnel` will likely need to be figured out together.

Thoughts?

zuiderkwast: Sounds good. I can do these changes.

For TCP errors, it's safest to close the socket IMO, not just the HTTP/2 frame. It's almost impossible to recover from a TCP error if the TCP layer hasn't already solved it.

(FWIW, we are using the simple patch which just sends a `tcp_error` message to self. It works well in practice.)

essen: It depends from where the error comes from. If it's from the real socket, then yes, everything should be closed. If it's from a fake socket (tunnels), only that tunnel should be closed. But it also depends on what protocol was used to setup the tunnel, as some don't allow subsequent requests (SOCKS) or parallel requests (HTTP, so it's probably safe to close). But HTTP/2 you can have both a tunnel and normal requests on the same connection so in that case we would only close the tunnel.

essen:
I have pushed a commit that handles HTTP/2 tunnel errors on top of the PR that was merged.

zuiderkwast:
> I have pushed a commit that handles HTTP/2 tunnel errors on top of the PR that was merged.

Awesome. I'll rebase this PR again onto latest master.

> In gun_http2, several functions may return a state or an error instead of only a state, e.g. frame/4 (...)

Do you think I should rewrite all these to return `{state, State} | {error, Reason}`?

essen:
I'm not sure, I was wondering about doing this with the tunnel changes but kept it simple. Considering the other changes you have in this PR (such as in gun_tunnel), perhaps this could be a good next step to normalize this in a separate PR. Let's retain the "StateOrError" variable names since they're not the full "Commands".

zuiderkwast:
I have updated this PR after #301 was merged. Is there anything else that should be done in a separate PR or is this one small enough now?

src/gun_tunnel.erl
```diff
@@ -449,6 +451,8 @@ ws_send(Frames, State=#tunnel_state{protocol=Proto, protocol_state=ProtoState},
 
 %% Internal.
 
+%% Returns an error on send errors, a state otherwise
+%% @todo Handle error command.
```
essen, line 455: We do?

src/gun_http.erl
```diff
@@ -632,6 +634,7 @@ send_request(State=#http_state{socket=Socket, transport=Transport, version=Versi
 		headers => Headers
 	},
 	EvHandlerState1 = EvHandler:request_start(RequestEvent, EvHandlerState0),
+	%% @todo Handle send errors.
```
essen, line 637: This sounds like the only remaining Transport:send that isn't handled. Let's handle it? Shouldn't be too difficult as this function is only called 3 times. We could do `SendResult = Transport:send` and then only change what is returned. Maybe return `{SendResult, Authority,.... EvHandlerState}` and handle it in headers/request/ws_upgrade functions. In the error condition we need to keep the CookieStore and EvHandlerState but can ignore the rest.

test/connection_SUITE.erl
```diff
@@ -0,0 +1,139 @@
...
+%%
+%% THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+%% WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+%% MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+%% ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+%% WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+%% ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+%% OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+-module(connection_SUITE).
```
essen, line 15: Let's call it `send_errors_SUITE`.

essen:
Almost there!

src/gun_tunnel.erl
zuiderkwast: Ah right, merge conflict remains probably.

zuiderkwast: In `gun_tunnel` there are still two of these:

```Erlang
			%% @todo Handle error result from Proto:init/4
			{ok, _, ProtoState} = Proto:init(ReplyTo, OriginSocket, OriginTransport,
				ProtoOpts#{stream_ref => StreamRef, tunnel_transport => tcp}),
```

essen: OK I will do those after merge. But the todo pointed to here can be removed.

src/gun_http.erl
zuiderkwast: Done.

test/connection_SUITE.erl
zuiderkwast: Done.

src/gun_tunnel.erl
zuiderkwast: I've removed the todo.

Now I've also solved the last remaining send error by allowing `gun_tunnel:init/6` to return an error and handle it where it's called. This is in the last commit. If you have something else in mind, I can revert it.

essen: OK I think that one is only partial. We don't want the entire connection to go down just because there's an error in the tunnel. The tunnel might be using a different socket than the connection (TLS proxy). But with the current code it's probably OK for now. I will just add a todo noting this to be improved later.

essen:
Merged, thanks!

PR #244, opened 2020-09-23
Normalize headers in ws_upgrade

rinpatch:
In the documentation headers passed to ws_upgrade are typed as
gun:req_headers(), however if a map of headers is passed (which is
allowed by the type), the gun process will crash when trying to operate
on it as if it were a list.


essen:
PR is fine but is missing a test. Note that I am not merging anything until the `http2-connect` branch is merged this or next week.

rinpatch:
Added a test that headers are normalized on ws_upgrade

essen:
Merged, thanks!

PR #245, opened 2020-09-29
[poc][don't merge] Check current max_concurrent_streams setting before attempting to open a stream.

bobergj:


bobergj:
Closing as the attempted solution is incorrect

PR #249, opened 2020-10-06
Export declared event types in gun_event

keynslug:
This should help library users in keeping event handler specs concise and correct.

One may consider these types a part of the _public API_ once they're exported though, maybe that's the reason they weren't?

### DIFF
--- a/src/gun_event.erl
+++ b/src/gun_event.erl
@@ -24,6 +24,7 @@
 	origin_port := inet:port_number(),
 	opts := gun:opts()
 }.
+-export_type([init_event/0]).
 
 -callback init(init_event(), State) -> State.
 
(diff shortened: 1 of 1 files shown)


essen:
As a rule if it's not documented then it's not public. The real reason is that it's a new feature in upcoming 2.0 and it hasn't been documented. The intent is that those types get reviewed and documented.

I would love to hear what you (or others) are doing with event handlers.

I will look into merging this tomorrow.

keynslug:
Ah, ok, great to hear!

> I would love to hear what you (or others) are doing with event handlers.

Well not much for the time being to be honest, besides occasional logging in tests. Worth mentioning that we have a strong requirement to have every public function typespec'ed in our projects, including gun event handlers, that's why I decided to open a PR for discussion. Later I'm planning to extract some additional log event metadata like, for example, domain lookup and request timings.

essen:
Merged, thanks!

Note that the event data may be subject to change before we reach 2.0, particularly as I will go through them in the context of HTTP/2 tunnels.

PR #250, opened 2020-10-19
Disconnect when setopts returns an error

zuiderkwast:
There is a race condition: After receiving data, before setting the socket back to {active, once}, the server may have closed the socket or an error may have occurred. Since the socket was in passive mode, no message is received. Handling the return value of setopts solves the issue.

----

This was discovered when investigating failing and unstable test cases in Cowboy. The cause was traced back to Gun (which is used for testing Cowboy). The Cowboy test case waits for `gun:await_body` to return `{error, {stream_error, closed}}` which never comes. Cowboy test cases which (sometimes or always) got stuck until timeout or time trap in this way, and solved by this fix, are these:

* `req_SUITE:stream_body_content_length_nofin_error/1` (only CT group is `https`)
* `rfc7231_SUITE:expect_discard_body_close/1` (only CT group is `https` and `https_compress`)

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -1780,8 +1780,14 @@ disconnect_flush(State=#state{socket=Socket, messages={OK, Closed, Error}}) ->
 active(State=#state{active=false}) ->
 	State;
 active(State=#state{socket=Socket, transport=Transport}) ->
-	Transport:setopts(Socket, [{active, once}]),
-	State.
+	case Transport:setopts(Socket, [{active, once}]) of
+		ok ->
+			State;
+		{error, closed} ->
+			disconnect(State, closed);
+		Error={error, _} ->
+			disconnect(State, Error)
+	end.
 
 status(State=#state{status={up, OwnerRef}}, NewStatus) ->
 	demonitor(OwnerRef, [flush]),


src/gun.erl
```diff
@@ -1780,8 +1780,12 @@ disconnect_flush(State=#state{socket=Socket, messages={OK, Closed, Error}}) ->
 active(State=#state{active=false}) ->
 	State;
 active(State=#state{socket=Socket, transport=Transport}) ->
-	Transport:setopts(Socket, [{active, once}]),
-	State.
+	case Transport:setopts(Socket, [{active, once}]) of
+		{error, closed} ->
```
essen, line 1784: What about other error conditions?

zuiderkwast: I didn't solve them in this PR. :-) I assume they can be handled in the same way, but I don't have any particular case right now.

zuiderkwast: Note that `closed` is not a network error. It's just that the server has closed the socket and we may have missed it while we were in passive mode.

essen: Sure but you can get a socket closed for other reasons so if we're going to handle one case we should handle all of them.

zuiderkwast: If it's as easy as adding another case `{error, _} = Error -> disconnect(State, Error)`, then sure. I checked now and it doesn't break any test cases at least. Would this be enough?

If this gets bigger, I think having no errors in the Cowboy test suites is quite nice to start with, so you can rely on the build results for PRs...

zuiderkwast: Done.

essen:
OK that can't possibly work because you are simply calling `disconnect/2` and this function returns a state change tuple, not state data. I will close this PR and do a better fix because I need to make Cowboy green in order to release it today. Huge thanks for figuring out where the problem comes from though!

zuiderkwast:
> OK that can't possibly work because you are simply calling `disconnect/2` and this function returns a state change tuple, not state data.

Oh, I didn't check that. :) I assumed it would exit and not return at all.

Simplest solution (one that you probably don't like) is to fake a message from the socket by doing `self() ! {ssl_closed, Socket}` from within active/1 and let gen_statem handle it.

essen:
No because you'll queue the message after all that are already waiting. I'm just handling the return value, it's only called 10 times.

PR #252, opened 2020-10-21
Fail fast when request is queued while closing if reconnect is off

zuiderkwast:
If a request is created when a connection is in state 'closing', e.g. after receiving an HTTP/2 GOAWAY frame or an HTTP/1.1 `Connection: close` header, an error message is sent back to the caller immediately, if reconnect is off (i.e. if the option retry is set to 0).

This allows an application to retry the request on another connection without waiting for all streams on the current connection to complete.

### DIFF
--- a/test/shutdown_SUITE.erl
+++ b/test/shutdown_SUITE.erl
@@ -218,9 +218,9 @@ http1_request_connection_close_pipeline(Config) ->
 	StreamRef3 = gun:get(ConnPid, "/"),
 	%% We get the response, pipelined streams get canceled, followed by Gun shutting down.
 	{response, nofin, 200, _} = gun:await(ConnPid, StreamRef1),
+	{error, {stream_error, closing}} = gun:await(ConnPid, StreamRef2),
+	{error, {stream_error, closing}} = gun:await(ConnPid, StreamRef3),
 	{ok, _} = gun:await_body(ConnPid, StreamRef1),
-	{error, {stream_error, {closed, normal}}} = gun:await(ConnPid, StreamRef2),
-	{error, {stream_error, {closed, normal}}} = gun:await(ConnPid, StreamRef3),
 	gun_is_down(ConnPid, ConnRef, normal).
 
 http1_response_connection_close(_) ->
@@ -272,8 +272,8 @@ http1_response_connection_close_pipeline(_) ->
 		%% We get the response, pipelined streams get canceled, followed by Gun shutting down.
 		{response, nofin, 200, _} = gun:await(ConnPid, StreamRef1),
 		{ok, _} = gun:await_body(ConnPid, StreamRef1),
-		{error, {stream_error, {closed, normal}}} = gun:await(ConnPid, StreamRef2),
-		{error, {stream_error, {closed, normal}}} = gun:await(ConnPid, StreamRef3),
+		{error, {stream_error, closing}} = gun:await(ConnPid, StreamRef2),
+		{error, {stream_error, closing}} = gun:await(ConnPid, StreamRef3),
 		gun_is_down(ConnPid, ConnRef, normal)
 	after
 		cowboy:stop_listener(?FUNCTION_NAME)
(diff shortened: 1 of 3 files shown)


essen:
You need to do this for `{headers...}`, `{connect...}` and `{ws_upgrade...}` messages as well.

zuiderkwast:
> You need to do this for `{headers...}`, `{connect...}` and `{ws_upgrade...}` messages as well.

Done.

essen:
I think this should be tested against HTTP/1.1 as well by sending a connection: close response that doesn't finish immediately.

zuiderkwast:
> I think this should be tested against HTTP/1.1 as well by sending a connection: close response that doesn't finish immediately.

Good idea. I'll do that. Implementation should already be in place, I hope.

essen:
Shouldn't be too difficult via Gun's gun_test origin functions or via Cowboy setting the connection: close header manually. Whichever fits best.

zuiderkwast:
I can add it in `shutdown_SUITE:http1_response_connection_close_pipeline/1` perhaps.... Just need to add a delay between the server response headers and the response body.

essen:
No make a new test. Delays create flaky tests.

zuiderkwast:
I've implemented this for HTTP/1.x as well now. Only a test case wasn't enough. I had to add some logic in `gun_http` where the headers are handled.

test/shutdown_SUITE.erl
```diff
@@ -468,7 +526,14 @@ http2_server_goaway_many_streams(_) ->
 	StreamRef1 = gun:get(ConnPid, "/"),
 	StreamRef2 = gun:get(ConnPid, "/"),
 	StreamRef3 = gun:get(ConnPid, "/"),
+	StreamRef4 = gun:get(ConnPid, "/"),
 	ConnRef = monitor(process, ConnPid),
+	%% GOAWAY received. Stream 4 is cancelled.
+	{gun_error, ConnPid, StreamRef4, Reason4} = receive E4 -> E4 end,
+	{goaway, no_error, _} = Reason4,
+	timer:sleep(100),
```
essen, line 534: I don't think this timer:sleep is necessary.

test/shutdown_SUITE.erl
```diff
@@ -456,7 +513,8 @@ http2_server_goaway_many_streams(_) ->
 		ok = Transport:send(Socket, [
 			cow_http2:headers(5, fin, HeadersBlock3)
 		]),
-		timer:sleep(500)
+		%% Gun closes the connection.
+		{error, closed} = gen_tcp:recv(Socket, 9, 500)
```
essen, line 517: Set timeout to `infinity`.

test/shutdown_SUITE.erl
```diff
@@ -297,6 +297,54 @@ http10_connection_close(Config) ->
...
+	%% Pipelined request fails immediately.
+	{gun_error, ConnPid, StreamRef2, closing} = receive E2 -> E2 end,
+
+	{gun_data, ConnPid, StreamRef1, nofin, <<"Hello">>} =
+		receive PartialBody -> PartialBody end,
+
+	%% Request initiated when Gun is in closing state fails immediately.
+	StreamRef3 = gun:get(ConnPid, "/"),
+	{gun_error, ConnPid, StreamRef3, closing} = receive E3 -> E3 end,
+
```
essen, line 343: Please remove all these empty lines while you're at it.

test/shutdown_SUITE.erl
zuiderkwast: Done (using recv/2).

test/shutdown_SUITE.erl
zuiderkwast: Done.

test/shutdown_SUITE.erl
zuiderkwast: You're right. Deleted.

essen:
Thanks, looking into merging this now. If you have further Gun stuff to submit or that I overlooked please send or ping the relevant ticket, I will be releasing 2.0.0-rc.1 next week.

zuiderkwast:
Great. I don't have anything more for Gun right now, but I'll let you know if I come up with something.

essen:
Merged, thanks!

filmor:
This breaks for me with an endpoint that implements Server Sent Events but includes a `Connection: close`. I have no idea whether this is allowed by a spec, but it definitely exists in the wild. Since there is no way to override this behaviour, I had to revert this patch to be able to receive more than one message from the server.

essen:
Please provide a reproducible test case in a separate ticket or PR.

filmor:
See #271.

PR #257, opened 2021-01-11
Add Gun pools

essen:
The approach taken here is very similar to what browsers are
doing. A separate pool is created for each host/port/scope.
The authority (host header) is used to determine which pool
will execute requests. A connection process is semi-randomly
chosen, from the connections that have capacity. Maximum
capacity is determined by the protocol (the HTTP/2 setting
set by the server is used, for example). Multiple processes
can process requests/responses on the same connection
concurrently. There is no need to "give back" the response
to the pool, the number of ongoing streams is maintained via
an event handler.

The implementation is currently not strict, there may be
more attempts to create requests than there is capacity.
I'm not sure if it should be made strict or if Gun should
just wait before sending requests (it only matters in the
HTTP/2 case at the moment).

When there is no connection with capacity available in the
pool (because they have too many streams, or are reconnecting,
or any other reason), checking out fails. There is no timeout
to wait for a connection to be available. On the other hand
the checkout_retry option allows setting multiple timeouts
to retry checking out a connection. Each retry attempt's
wait time can have a different value.

The initial implementation of this work was sponsored by
Kobil and made at the suggestion of Ilya Khaprov.

### DIFF
--- a/src/gun_pool.erl
+++ b/src/gun_pool.erl
@@ -1,0 +1,704 @@
+%% Copyright (c) 2021, Loïc Hoguin <essen@ninenines.eu>
+%%
+%% Permission to use, copy, modify, and/or distribute this software for any
+%% purpose with or without fee is hereby granted, provided that the above
+%% copyright notice and this permission notice appear in all copies.
+%%
+%% THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+%% WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+%% MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+%% ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+%% WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+%% ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+%% OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+-module(gun_pool).
+-behaviour(gen_statem).
+
+%% Pools.
+-export([start_pool/3]).
+-export([stop_pool/2]).
+-export([stop_pool/3]).
+% @todo shutdown pool?
+-export([info/0]).
+-export([info/1]).
+-export([info/2]).
+-export([await_up/1]).
+-export([await_up/2]).
+-export([checkout/2]). %% Use responsibly!
+
+%% Requests.
+-export([delete/2]).
+-export([delete/3]).
+-export([get/2]).
+-export([get/3]).
+-export([head/2]).
+-export([head/3]).
+-export([options/2]).
+-export([options/3]).
+-export([patch/2]).
+-export([patch/3]).
+-export([patch/4]).
+-export([post/2]).
+-export([post/3]).
+-export([post/4]).
+-export([put/2]).
+-export([put/3]).
+-export([put/4]).
+
+%% Generic requests interface.
+-export([headers/3]).
+-export([headers/4]).
+-export([request/4]).
+-export([request/5]).
+
+%% Streaming data.
+-export([data/3]).
+
+%% Tunneling. (HTTP/2+ only.)
+%% @todo -export([connect/2]).
+%% @todo -export([connect/3]).
+%% @todo -export([connect/4]).
+
+%% Cookies.
+%% @todo -export([gc_cookies/1]).
+%% @todo -export([session_gc_cookies/1]).
+
+%% Awaiting gun messages.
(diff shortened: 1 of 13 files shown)


src/gun_pool.erl
```diff
@@ -0,0 +1,600 @@
...
+		ManagerPid ->
+			case get_connection(ManagerPid, WsSendOpts) of
+				undefined ->
+					error; %% @todo
+				{ConnPid, #{ws := StreamRef}} ->
+					gun:ws_send(ConnPid, StreamRef, Frames)
+			end
+	end.
+
+%% Pool manager internals.
```
essen, line 377: More detailed explanations start here.

src/gun_pool.erl
```diff
@@ -0,0 +1,600 @@
...
+
+start_link(Host, Port, Opts) ->
+	gen_statem:start_link(?MODULE, {Host, Port, Opts}, []).
+
+%% @todo Probably don't mix Gun and pool options. Have Gun Opts in a pool option.
+init({Host, Port, Opts}) ->
+	Transport = maps:get(transport, Opts, default_transport(Port)),
+	Authority = gun_http:host_header(Transport, Host, Port),
+	Scope = maps:get(scope, Opts, default),
+	true = ets:insert_new(gun_pools, {{Scope, iolist_to_binary(Authority)}, self()}),
```
essen, line 419: How scope works. Nothing fancy here.

src/gun_pool.erl
```diff
@@ -0,0 +1,600 @@
...
+handle_common({call, From}, {get_connection, _ReqOpts}, _,
+		StateData=#state{conns_meta=ConnsMeta}) ->
+	case find_available_connection(StateData) of
+		none ->
+			{keep_state_and_data, {reply, From, undefined}};
+		ConnPid ->
+			Meta = maps:get(ConnPid, ConnsMeta, #{}),
+			{keep_state_and_data, {reply, From, {ConnPid, Meta}}}
+	end;
+handle_common(info, {gun_notify, ConnPid, settings_changed, Settings}, _, StateData=#state{conns=Conns}) ->
```
essen, line 525: We receive a message when HTTP/2 settings change. We keep track of the current max_concurrent_streams value this way.

src/gun_pool.erl
```diff
@@ -0,0 +1,600 @@
...
+	handle_common(Type, Event, ?FUNCTION_NAME, StateData).
+
+setup_fun(#{setup_fun := SetupFun}) ->
+	SetupFun;
+setup_fun(_) ->
+	{fun (_, {gun_up, _, Protocol}, _) ->
+		{up, Protocol, #{}}
+	end, undefined}.
+
+degraded_setup(ConnPid, Msg, StateData0=#state{conns=Conns, conns_meta=ConnsMeta}, SetupFun, SetupState0) ->
```
essen, line 488: An optional setup fun can be used to reestablish the Websocket connection (or later, tunnel).

src/gun_pool.erl
```diff
@@ -0,0 +1,600 @@
...
+	keep_state_and_data.
+
+%% We go over every connection and return the first one
+%% we find that has capacity. How we determine whether
+%% capacity is available depends on the protocol. For
+%% HTTP/2 we look into the protocol settings. The
+%% current number of streams is maintained by the
+%% event handler gun_pool_events_h.
+find_available_connection(#state{table=Tid, conns=Conns}) ->
+	I = lists:sort([{rand:uniform(), K} || K <- maps:keys(Conns)]),
```
essen, line 564: How randomness is added in connection selection.

src/gun_pool_events_h.erl
```diff
@@ -0,0 +1,157 @@
...
+	propagate(Event, State, ?FUNCTION_NAME).
+
+tls_handshake_start(Event, State) ->
+	propagate(Event, State, ?FUNCTION_NAME).
+
+tls_handshake_end(Event, State) ->
+	propagate(Event, State, ?FUNCTION_NAME).
+
+request_start(Event=#{stream_ref := StreamRef}, State=#{table := Tid}) ->
+	_ = ets:update_counter(Tid, self(), +1, {self(), 0}),
```
essen, line 68: This module will +1 or -1 the current stream count based on events. This is why the pool is not strict at the moment, this occurs after the pid has been given to the calling process.

test/pool_SUITE.erl
```diff
@@ -0,0 +1,311 @@
...
+	Streams = [gun_pool:get("/",
+		#{<<"host">> => ["localhost:", integer_to_binary(Port)]},
+		#{scope => ?FUNCTION_NAME}
+	) || _ <- lists:seq(1, 800)],
+	_ = [begin
+		{response, nofin, 200, _} = gun_pool:await(StreamRef),
+		{ok, <<"Hello world!">>} = gun_pool:await_body(StreamRef)
+	end || StreamRef <- Streams].
+
+hello_pool_ws(Config) ->
```
essen, line 88: You can also use it with Websocket but it requires the use of a Websocket handler (above) to forward messages to the correct process.

essen:
I think as far as the interface goes, this is ready. Needs more tests and documentation of course but I'm not sure I want to document this for 2.0, maybe 2.1.

essen:
Merged.

PR #264, opened 2021-03-12
Fixed problem with multiple server_name_indication in tls opts. Havin…

jbevemyr:
…g multiple server_name_indication entries triggers a problem in OTP 23.2.7.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
 {deps, [
-{cowlib,".*",{git,"https://github.com/ninenines/cowlib","2.10.1"}}
+{cowlib,".*",{git,"https://github.com/ninenines/cowlib","master"}}
 ]}.
 {erl_opts, [debug_info,warn_export_vars,warn_shadow_vars,warn_obsolete_guard]}.
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -141,6 +141,7 @@
 	http_opts => http_opts(),
 	http2_opts => http2_opts(),
 	protocols => protocols(),
+	raw_opts => raw_opts(),
 	retry => non_neg_integer(),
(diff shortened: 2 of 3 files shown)


essen:
How do you use Gun, that you have to set SNI manually?

jbevemyr:
We use gun to talk to a server that doesn't have a hostname that matches the certificate, so instead we set the SNI.
We have a cluster of nodes where each node runs a cowboy server with the same certificate. We can connect to them with different host names but still verify them with the SNI. A workaround would be to issue separate certificates for each host and have dns alt names, of course.

jbevemyr:
Same fix, just slightly differently coded.

michaelklishin:
This looks reasonable to me. Most users won't need it but the risk of accepting this PR seem low.

essen:
Sure. Could you please add a test though? Something similar to https://github.com/ninenines/gun/blob/master/test/gun_SUITE.erl#L81 except you'll want an event that occurs after `ensure_alpn_sni`.

essen:
It appears that since https://github.com/erlang/otp/commit/e9b0dbb4a95 the `ssl` application already sets SNI if none is provided. Therefore I think the best course of action here is to simply remove this code. I will add a test to confirm that SNI is sent and that a custom SNI can be set, regardless.

essen:
OK the automatic SNI applies to directly connecting using `ssl`. It does not apply to upgrading a TCP socket. For that we must specify SNI (there's a note about this in https://www.erlang.org/doc/man/ssl.html#connect-3). So the patch is still necessary.

essen:
Cherry-picked the relevant commit. Thanks!

The tests I've added can't catch the original issue, but they will be good to have for the future.

PR #266, opened 2021-04-19
handle any zero-sized Data in http:data with fin

mbj4668:
I passed `[]` as Data to gun:data/4 with `fin`, and got really weird results...
Eventually I found this.

### DIFF
--- a/test/gun_SUITE.erl
+++ b/test/gun_SUITE.erl
@@ -107,6 +107,21 @@ ignore_empty_data_http(_) ->
 	1 = length(Zero),
 	gun:close(Pid).
 
+ignore_empty_data_fin_http(_) ->
+	doc("When gun:data/4 is called with fin and empty data, it must be ignored."),
+	{ok, OriginPid, OriginPort} = init_origin(tcp, http),
+	{ok, Pid} = gun:open("localhost", OriginPort),
+	{ok, http} = gun:await_up(Pid),
+	handshake_completed = receive_from(OriginPid),
+	Ref = gun:post(Pid, "/", []),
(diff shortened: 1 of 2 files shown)


essen:
Please add a test if possible.

test/gun_SUITE.erl
```diff
@@ -107,6 +107,21 @@ ignore_empty_data_http(_) ->
...
+	{ok, Pid} = gun:open("localhost", OriginPort),
+	{ok, http} = gun:await_up(Pid),
+	handshake_completed = receive_from(OriginPid),
+	Ref = gun:post(Pid, "/", []),
+	gun:data(Pid, Ref, nofin, "hello"),
+	gun:data(Pid, Ref, fin, ""),
+	Data = receive_all_from(OriginPid, 500),
+	Lines = binary:split(Data, <<"\r\n">>, [global]),
+	Zero = [Z || <<"0">> = Z <- Lines],
+	1 = length(Zero),
```
essen, line 122: Took me a while to figure out you were looking for 0\r\n. Maybe search for `<<"\r\n0\r\n">>` via `binary:match` instead.

mbj4668: I copied the existing test case `ignore_empty_data_http`.
W/o the fix, the code sent two "0".  This test ensures it just gets one.

essen: Even better if you can fix that test as well.

essen: Perhaps instead of doing this it would be better to split in two at `\r\n\r\n` and try to parse the body.

Payaspx (approved):

PR #271, opened 2021-07-02
Ignore parameters for text/event-stream

filmor:


essen:
Thanks. Please add a test to https://github.com/ninenines/gun/blob/master/test/sse_SUITE.erl

filmor:
I have added a test for the MIME type parameter as well on the issue with `Connection: close` that I reported in #252.

The latter is failing right now.

filmor:
@essen Would you need anything else to merge this one?

essen:
Time flies. Sorry, will add it to the top of the pile.

src/gun_sse_h.erl
```diff
@@ -25,15 +25,19 @@
...
 
 -spec init(pid(), reference(), _, cow_http:headers(), _)
 	-> {ok, #state{}} | disable.
 init(ReplyTo, StreamRef, _, Headers, _) ->
 	case lists:keyfind(<<"content-type">>, 1, Headers) of
-		{_, <<"text/event-stream">>} ->
-			{ok, #state{reply_to=ReplyTo, stream_ref=StreamRef,
-				sse_state=cow_sse:init()}};
+		{_, ContentType} ->
+			case cow_http_hd:parse_content_type(ContentType) of
```
essen, line 34: I will merge it now but if you want to another test to add would be sending back a content-type that's garbage (for example `content-type: garbage`). I believe it will result in a crash right now. It probably should just return `disable` instead.

essen:
Merged, thanks!

Issue #273, opened 2021-07-22
Flushing informational responses by stream reference


Yozhig:
https://github.com/ninenines/gun/blob/f9175998687678e227bdd49669e2d83f0648fa57/src/gun.erl#L877
Shouldn't there be a call to `flush_ref/1` instead of `flush_pid/1`?

essen:
Yes.

essen:
Closing in favor of the PR. Thanks!

PR #274, opened 2021-07-22
Fix flushing by stream reference

Yozhig:
I've tried to figure out why dialyzer is silent about this after adding specs, so fixed few specs for other functions as well.

Fixes ninenines/gun#273


essen:
Thanks! I have removed some of the changes (keeping things minimal for now with regard to Dialyzer) and merged.

PR #276, opened 2021-08-10
implement gun_raw:down/1, gun_raw:update_flow/4

chilikk:
Avoid crash (undefined gun_raw:down) on TCP connection being closed when the connection is handled by gun_raw by implementing a rudimentary gun_raw:down/1.

Implement flow control in gun_raw by adding a flow option in raw_opts and implementing gun_raw:update_flow/4


essen:
Hello, thanks for the patch. Please add tests so we know it works.

essen:
Merged, thanks!

PR #277, opened 2021-08-30
Accept HTTP/1.0 response to CONNECT

krant:
There are still proxies in a wild which respond `HTTP/1.0 200` to `CONNECT`. With this PR Gun would support those proxies.

### DIFF
--- a/src/gun_http.erl
+++ b/src/gun_http.erl
@@ -310,7 +310,7 @@ handle_head(Data, State=#http_state{opts=Opts,
 
 handle_connect(Rest, State=#http_state{
 		streams=[Stream=#stream{ref={_, StreamRef, Destination}, reply_to=ReplyTo}|Tail]},
-		CookieStore, EvHandler, EvHandlerState0, 'HTTP/1.1', Status, Headers) ->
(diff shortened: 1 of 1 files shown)


essen:
Please be more specific. Which proxies?

krant:
For example squidproxies.com still sells squid/3.1.23, you could check it with one of IP we bought from them:
```
curl -kvx 23.19.34.138:8800 https://httpbin.org/ip
```
(it will give you an error because of authentication, but `HTTP/1.0` and squid version is there)

essen:
Your `curl` command is something else. It's fine to return an HTTP/1.0 for a non-2xx response to a CONNECT request. Gun already handle those cases properly I believe.

It is the HTTP/1.0 2xx responses to CONNECT specifically that are in question here.

In any case I cannot think of a reason why this would create problems. On the other hand it would be good to add a comment explaining why we also accept HTTP/1.0 and that it is unlikely to cause problems because the tunnel takes over the entire connection.

Also one or more tests to confirm that the code behaves as intended.

essen:
This has been fixed today. Closing, thanks!

PR #280, opened 2021-10-28
Respect remote concurrency limit

zuiderkwast:
If the number of streams has reached the server setting MAX_CONCURRENT_STREAMS, new requests result in a stream error `too_many_streams`.

This allows the user to handle the situation, e.g. to retry the request later or to send it on a different connection.

Depends on: ninenines/cowlib#123

### DIFF
--- a/test/rfc7540_SUITE.erl
+++ b/test/rfc7540_SUITE.erl
@@ -392,6 +392,7 @@ respect_max_concurrent_streams(_) ->
 		{ok, ConnPid} = gun:open("localhost", Port, #{protocols => [http2]}),
 		{ok, http2} = gun:await_up(ConnPid),
 		StreamRef1 = gun:get(ConnPid, "/delayed"),
+		timer:sleep(100),
 		StreamRef2 = gun:get(ConnPid, "/delayed"),
 		{error, {stream_error, Reason}} = gun:await(ConnPid, StreamRef2),
 		{stream_error, too_many_streams, _Human} = Reason,


essen:
I agree the retry logic should be done in the client because it is the most adequate to decide what to do about it. Some requests must go through whatever the cost, some can be dropped. Backoff strategies may differ. Gun usage with sync/async may differ.

zuiderkwast:
> 

@essen so you agree with this PR, returning an error? Then gun respects the RFC and doesn't go over the limit.

What's left to do AFAICT is to the get the corresponding PR for cowlib merged, use the upstream cowlib again (and remove some debug stuff I accidentally committed).

essen:
I need to think a little more about it. I think the error returned should be specific, not `badstate`, so users can react as appropriate. We can probably use `too_many_streams` or something to that effect. I'm not sure yet about the Cowlib part. I will go through the other stuff marked as 2.0 (Gun 2.0 in Cowlib) before coming back to this.

zuiderkwast:
Changed the error to `too_many_streams`.

zuiderkwast:
Bump. This one in 2.0?

essen:
If I have time. I want 2.0 to be out before next year so I am focusing on what potentially break the interface.

zuiderkwast:
Ping. This one is ready for review too.

The merge conflict is for the temporary dependency of a cowlib branch which this PR depends on, in Makefile and rebar.config, e.g.

```
<<<<<<< HEAD
dep_cowlib = git https://github.com/Nordix/cowlib respect-remote-concurrency-limit
=======
dep_cowlib = git https://github.com/ninenines/cowlib 2.12.1
>>>>>>> origin/master
```

No point resolving before the cowlib PR is merged, right?

essen:
No need. I'm planning to go over the PRs later this week.

test/rfc7540_SUITE.erl
```diff
@@ -375,6 +375,36 @@ lingering_data_counts_toward_connection_window(_) ->
...
 
+respect_max_concurrent_streams(_) ->
+	doc("The SETTINGS_MAX_CONCURRENT_STREAMS setting can be used to "
+		"restrict the number of concurrent streams. (RFC7540 5.1.2, RFC7540 6.5.2)"),
+	{ok, Ref, Port} = do_cowboy_max_councurrent_streams(1),
+	try
+		{ok, ConnPid} = gun:open("localhost", Port, #{protocols => [http2]}),
+		{ok, http2} = gun:await_up(ConnPid),
+		StreamRef1 = gun:get(ConnPid, "/"),
+		timer:sleep(100),
```
essen, line 386: Probably don't need the timer:sleep here.

test/rfc7540_SUITE.erl
```diff
@@ -375,6 +375,36 @@ lingering_data_counts_toward_connection_window(_) ->
...
+		{response, nofin, 200, _} = gun:await(ConnPid, StreamRef1),
+		{ok, _} = gun:await_body(ConnPid, StreamRef1),
+		gun:close(ConnPid)
+	after
+		cowboy:stop_listener(Ref)
+	end.
+
+do_cowboy_max_councurrent_streams(MaxConcurrentStreams) ->
+	Ref = make_ref(),
+	Routes = [{'_', [{"/", delayed_hello_h, 500}]}],
```
essen, line 399: Let's change `"/"` to `"/delayed"` to make it more obvious.

Also we probably don't need to do the Cowboy init in a separate function, we can do it in the test directly. The other function `do_cowboy_origin` is used in multiple places that's why it's separate.

src/gun_http2.erl
```diff
@@ -970,7 +970,43 @@ headers(State, RealStreamRef=[StreamRef|_], ReplyTo, Method, _Host, _Port,
...
 
-request(State0=#http2_state{socket=Socket, transport=Transport, opts=Opts,
+request(State=#http2_state{http2_machine=HTTP2Machine}, StreamRef, ReplyTo, Method, Host, Port,
+		Path, Headers, Body, InitialFlow, CookieStore, EvHandler, EvHandlerState)
+		when is_reference(StreamRef) ->
+
+	case cow_http2_machine:is_remote_concurrency_limit_reached(HTTP2Machine) of
+		true ->
+			ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef),
+				too_many_streams},
```
essen, line 980: Instead of `too_many_streams` it should be `{stream_error, too_many_streams, 'Maximum concurrency limit has been reached.'}` or similar.

test/rfc7540_SUITE.erl
zuiderkwast: Done.

zuiderkwast:
Rebased.

src/gun_http2.erl
zuiderkwast: Done.

test/rfc7540_SUITE.erl
zuiderkwast: Done.

belowEV:
Hej @essen,
I apologize for stressing the worst question but maybe you have some vision when this PR could be included in gun release ?
Thanks in advance!

essen:
I will go over pending PRs and tickets in the coming weeks now that the HTTP/3 work has been merged. I would like to merge most of them. I don't have a timeline for a release at this time though.

Gun has a few CI failures that would be good to address before then, if you're looking for something to do, see my comment at https://github.com/ninenines/gun/pull/319#issuecomment-2020365739

This and other PRs will likely need a rebase as well.

test/rfc7540_SUITE.erl
```diff
@@ -377,6 +377,31 @@ lingering_data_counts_toward_connection_window(_) ->
...
+	ProtoOpts = #{
+		env => #{dispatch => cowboy_router:compile(Routes)},
+		tcp => #{protocols => [http2]},
+		max_concurrent_streams => 1
+	},
+	[{ref, _}, {port, Port}] = gun_test:init_cowboy_tcp(Ref, ProtoOpts, []),
+	try
+		{ok, ConnPid} = gun:open("localhost", Port, #{protocols => [http2]}),
+		{ok, http2} = gun:await_up(ConnPid),
+		StreamRef1 = gun:get(ConnPid, "/delayed"),
```
bjosv, line 394: There is an intermittent timing issue in this testcase, a remedy would be:  
```suggestion
		StreamRef1 = gun:get(ConnPid, "/delayed"),
		timer:sleep(100),
```

essen:
I will be away for a week. If you can update in the meantime, great! For example Cowlib 2.14 has the needed function so should be updated. If you can't, no worries, I'll get back to it then.

bjosv:
I have updated it now.

test/rfc7540_SUITE.erl
bjosv: Ah, yes great. I mostly got this problem when running on a loaded machine.
I guess you have your own CI setup, but it would most likely be seen on Githubs CI machines..

essen: It's because we have to wait for settings to be received by Gun before the limit can be applied. Since we now have `notify_settings_changed`, I enabled it for the test and the issue went away. There's more to do on this PR but I'm on it.

essen:
I have opened https://github.com/ninenines/gun/pull/347 containing the missing parts of this PR.

PR #285, opened 2022-03-07
Return commands instead of state

zuiderkwast:
A command is returned instead of a state by these Protocol functions:

    headers/12
    request/13
    keepalive/3
    data/7
    connect/9
    ws_upgrade/11
    cancel/5

### DIFF
--- a/src/gun_http.erl
+++ b/src/gun_http.erl
@@ -546,36 +546,36 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
 	close_streams(State, Tail, Reason).
 
 %% We don't send a keep-alive when a CONNECT request was initiated.
-keepalive(State=#http_state{streams=[#stream{ref={connect, _, _}}]}, _, EvHandlerState) ->
-	{State, EvHandlerState};
+keepalive(#http_state{streams=[#stream{ref={connect, _, _}}]}, _, EvHandlerState) ->
+	{[], EvHandlerState};
 %% We can only keep-alive by sending an empty line in-between streams.
-keepalive(State=#http_state{socket=Socket, transport=Transport, out=head}, _, EvHandlerState) ->
+keepalive(#http_state{socket=Socket, transport=Transport, out=head}, _, EvHandlerState) ->
 	Transport:send(Socket, <<"\r\n">>),
-	{State, EvHandlerState};
-keepalive(State, _, EvHandlerState) ->
-	{State, EvHandlerState}.
+	{[], EvHandlerState};
+keepalive(_State, _, EvHandlerState) ->
+	{[], EvHandlerState}.
 
 headers(State, StreamRef, ReplyTo, _, _, _, _, _, _, CookieStore, _, EvHandlerState)
 		when is_list(StreamRef) ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef),
 		{badstate, "The stream is not a tunnel."}},
-	{State, CookieStore, EvHandlerState};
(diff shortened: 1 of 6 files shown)


essen:
Thanks for going through the hoops! I plan to review/merge tomorrow.

src/gun_http.erl
```diff
@@ -831,12 +831,12 @@ down(#http_state{streams=Streams}) ->
...
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef), {badstate,
 		"The stream has already been closed."}},
-	State.
+	{state, State}.
 
 error_stream_not_found(State, StreamRef, ReplyTo) ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef), {badstate,
 		"The stream cannot be found."}},
-	State.
+	{state, State}.
```
essen, line 839: I will change these two functions to return `ok` and move the calls out in order to return `[]` instead. Will be clearer.

src/gun_http.erl
```diff
@@ -546,45 +546,45 @@ close_streams(State, [#stream{ref=StreamRef, reply_to=ReplyTo}|Tail], Reason) ->
...
+	{Authority, Conn, Out, CookieStore, EvHandlerState} = send_request(State0,
 		StreamRef, ReplyTo, Method, Host, Port, Path, Headers, undefined,
 		CookieStore0, EvHandler, EvHandlerState0, ?FUNCTION_NAME),
 	InitialFlow = initial_flow(InitialFlow0, Opts),
-	{new_stream(State#http_state{connection=Conn, out=Out}, StreamRef, ReplyTo,
-		Method, Authority, Path, InitialFlow),
-		CookieStore, EvHandlerState}.
+	State = new_stream(State0#http_state{connection=Conn, out=Out}, StreamRef,
+		ReplyTo, Method, Authority, Path, InitialFlow),
+	{{state, State}, CookieStore, EvHandlerState}.
```
essen, line 572: OK the rest of the PR didn't introduce a `State` so I will remove it.

src/gun_http2.erl
```diff
@@ -1399,12 +1407,12 @@ connection_error(#http2_state{socket=Socket, transport=Transport,
...
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef), {badstate,
 		"The stream has already been closed."}},
-	State.
+	{state, State}.
 
 error_stream_not_found(State, StreamRef, ReplyTo) ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef), {badstate,
 		"The stream cannot be found."}},
-	State.
+	{state, State}.
```
essen, line 1415: Same as HTTP/1.1 for these functions.

essen (approved):
I will just do a minor tweak when merging and then I will change the error_stream_* functions.

src/gun_http.erl
essen: Sure please do I'm eating breakfast.

zuiderkwast: Done. Btw, have you considered using GitHub's "Squash merge" button? IMO it's practical to keep multiple commits within a PR just in case you need to split partial changes into a separate PR. ;-)

essen: Thanks! I'm on it.

I don't merge via GitHub's UI. The upstream repository is on my laptop, not on GitHub. I rebase and amend when necessary when I merge. I will probably keep the stream_* functions change in a separate commit anyway because it is a different change.

essen:
Merged, thanks!

If the original PR has something that can be added independently, please open a separate PR. Otherwise please rebase the PR and let's work on merging it!

PR #290, opened 2022-05-05
Fix infinite loop when connecting from a source in another subnet

juantascon:
There's an infinite loop when trying to connect to an ip if you provide a source ip address in another subnet, this can be reproduced like this:
`
:gun_pool.start_pool({20.20.20.1}, 8888, %{conn_opts: %{protocols: [:http2],tcp_opts: [ip: {127, 0, 0, 1}]}})
`

### DIFF
--- a/src/gun_tcp.erl
+++ b/src/gun_tcp.erl
@@ -76,7 +76,7 @@ connect(#{ip_addresses := IPs, port := Port, tcp_module := Mod, tcp_opts := Opts
 	end,
 	case Res of
 		{ok, S} -> {ok, S};
-		Error -> maybe_exit(Error)
+		Error -> Error
 	end.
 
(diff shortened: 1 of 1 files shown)


essen:
Thanks I will take a look soon.

juantascon:
@essen please do not forget to review

essen:
Please don't bump I will review when time allows.

essen:
OK I can reproduce. The fix is invalid because it only "fixes" `gun_pool` and not plain Gun usage. The right fix would be to do what this todo talks about: https://github.com/ninenines/gun/blob/master/src/gun_pools_sup.erl#L33

I have opened #292 to track this. Thanks!

essen:
This is fixed in the most recent commit. The pool will be left in a degraded state when this happens, but it will no longer try to open new Gun connections infinitely.

PR #293, opened 2022-05-12
Add state (current state name) info field in gun:info/1

zuiderkwast:
It's useful to be able to check the HTTP connection status (connected, closing, not connected, etc.). The current gen_statem state name provides exactly this info. This commit adds it to `gun:info/1` under the key `state`.

### DIFF
--- a/doc/src/manual/gun.info.asciidoc
+++ b/doc/src/manual/gun.info.asciidoc
@@ -23,7 +23,7 @@ Info :: #{
     origin_port    => inet:port_number(),
     intermediaries => [Intermediary],
     cookie_store   => gun_cookies:cookie_store(),
-    state          => atom()
+    state_name     => atom()
 }
 Intermediary :: #{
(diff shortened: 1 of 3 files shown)


zuiderkwast:
This one is waiting as well.

If you don't accept exposing the internal gen_statem states, then how about a variant with fewer states (e.g. `connecting | connected | closing`) instead of the full `not_connected | domain_lookup | connecting | initial_tls_handshake | tls_handshake | connected | closing`?

Or if you don't accept it at all, it's better to know that and close it.

essen:
Probably best to document it as `atom()` so they can be changed without it being a breaking change if necessary.

essen:
One final change, sorry I didn't think about this earlier. I think we should name the field `state_name` instead of `state` to avoid any unnecessary confusion with the process state.

essen (approved):

essen:
Merged, thanks!

Issue #254, opened 2020-12-03
Capture Ping ack response and Reply to Parent Process


sanjuktaindia:
Hi,
We would like to send the ping ack response to parent  process. So that we can monitor successive pings and raise flag for connection fail. Please provide suggestions how we can achieve it?

zuiderkwast:
Hi Loic,

we have a request from one of our product teams (a major Ericsson product) for a way to brutally close the connection if no ping ack (HTTP/2) is received after 3 pings. Currently, AFAICT ping_ack is just ignored, but I may have missed something.

I can imagine a configuration like `keepalive_tolerance => NumPingsWithoutAck :: non_neg_integer()`. This, together with the existing keepalive (milliseconds) option let's the user control this in a sufficiently granular way, I think.

Another way is by notifying the application about `ping_ack_received` as @sanjuktaindia suggested. In that case, we'd also need `ping_sent` to be sent to the application so it can keep track of when the ping is sent, as you commented. This gives full control to the application.

What do you think?

essen:
When you say after 3 pings, does that mean the connection gets closed when it is time to send a 4th ping and no ping_ack has been received? In that case I can see the configuration working out. It doesn't supersede the `idle_timeout` option though since that one would be protocol-agnostic since it's at the socket level.

About notifying the application, I believe it is something that, if necessary, should be done via the event handlers, perhaps with a new callback notifying both ping and ping acks.

Whichever of the two options you decide on let's do that in a separate ticket/PR.

JeppeTh:
Idle_timeout is rather about disconnecting idle connections (in cowboy at least)?

To determine if a non-idle connection is broken you don't need ping? Any request sent on the non-idle connection can be used for that?

essen:
There's no real difference between a request and pings, as far as detecting broken connections go. If data goes in and out, the connection is working. Counting time between ping/ping_ack is the same as request/response, if you issue a request and don't get a response in time, you can drop the connection, you don't need to do a separate ping for that.

What ping does that everything else doesn't is help maintain connections open when the connection is otherwise idle.

essen:
This was merged. Thanks!

PR #294, opened 2022-05-13
Add keepalive_tolerance http2 option

zuiderkwast:
`keepalive_tolerance`: The number of unacknowledged pings that can be tolerated
before the connection is forcefully closed.

When a keepalive ping is sent to the peer, a counter is incremented and if this
counter exceeds the tolerance limit, the connection is forcefully closed. The
counter is reset whenever a ping ack is received from the peer.

By default, the mechanism for closing the connection based on ping and ping ack
is disabled.

As discussed in #254.

### DIFF
--- a/doc/src/manual/gun.asciidoc
+++ b/doc/src/manual/gun.asciidoc
@@ -209,6 +209,7 @@ http2_opts() :: #{
     cookie_ignore_informational => boolean(),
     flow                        => pos_integer(),
     keepalive                   => timeout(),
+    keepalive_tolerance         => non_neg_integer(),
 
     %% HTTP/2 state machine configuration.
     connection_window_margin_size  => 0..16#7fffffff,
@@ -256,6 +257,19 @@ By default flow control is disabled.
 keepalive (infinity)::
 
 Time between pings in milliseconds.
+This option by itself does not cause the connection to be closed.
+To force a connection close if no ping ack is received before sending more
+pings, use `keepalive_tolerance` together with `keepalive`.
+
+keepalive_tolerance::
+
+The number of unacknowledged pings that can be tolerated before the connection
+is forcefully closed.
+When a keepalive ping is sent to the peer, a counter is incremented and if this
+counter exceeds the tolerance limit, the connection is forcefully closed. The
+counter is decremented when a ping ack is received from the peer.
+By default, the mechanism for closing the connection based on ping and ping ack
+is disabled.
 
 === opts()
 
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -89,6 +89,7 @@
 	opts = #{} :: gun:http2_opts(),
 	content_handlers :: gun_content_handler:opt(),
 	buffer = <<>> :: binary(),
+	pings_sent = 0 :: non_neg_integer(),
 
 	%% Base stream ref, defined when the protocol runs
 	%% inside an HTTP/2 CONNECT stream.
@@ -138,6 +139,8 @@ do_check_options([{flow, InitialFlow}|Opts]) when is_integer(InitialFlow), Initi
 do_check_options([{keepalive, infinity}|Opts]) ->
 	do_check_options(Opts);
 do_check_options([{keepalive, K}|Opts]) when is_integer(K), K > 0 ->
+	do_check_options(Opts);
+do_check_options([{keepalive_tolerance, K}|Opts]) when is_integer(K), K >= 0 ->
 	do_check_options(Opts);
 do_check_options([{notify_settings_changed, B}|Opts]) when is_boolean(B) ->
 	do_check_options(Opts);
@@ -341,7 +344,8 @@ frame(State=#http2_state{http2_machine=HTTP2Machine0}, Frame, CookieStore, EvHan
 	end.
 
 maybe_ack_or_notify(State=#http2_state{reply_to=ReplyTo, socket=Socket,
-		transport=Transport, opts=Opts, http2_machine=HTTP2Machine}, Frame) ->
+		transport=Transport, opts=Opts, http2_machine=HTTP2Machine,
+		pings_sent=PingsSent}, Frame) ->
 	case Frame of
 		{settings, _} ->
 			%% We notify remote settings changes only if the user requested it.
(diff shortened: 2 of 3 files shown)


zuiderkwast:
Test not yet added.

Please check if you think this feature makes sense. Then I'll add a test case.

src/gun_http2.erl
```diff
@@ -341,13 +344,15 @@ maybe_ack_or_notify(State=#http2_state{reply_to=ReplyTo, socket=Socket,
 				_ ->
 					ok
 			end,
-			Transport:send(Socket, cow_http2:settings_ack());
+			Transport:send(Socket, cow_http2:settings_ack()),
+			State;
 		{ping, Opaque} ->
 			Transport:send(Socket, cow_http2:ping_ack(Opaque));
+		{ping_ack, _Opaque} ->
+			State#http2_state{pings_sent = 0};
```
essen, line 352: That's probably wrong, it should be Current-1.

doc/src/manual/gun.asciidoc
```diff
@@ -257,6 +258,16 @@ keepalive (infinity)::
...
 
+keepalive_tolerance::
+
+The number of unacknowledged pings that can be tolerated before the connection
+is forcefully closed.
+When a keepalive ping is sent to the peer, a counter is incremented and if this
+counter exceeds the tolerance limit, the connection is forcefully closed. The
+counter is reset whenever a ping ack is received from the peer.
+By default, the mechanism for closing the connection based on ping and ping ack
+is disabled.
```
essen, line 269: We should probably enable it by default, but since *keepalives* are disabled by default, it'll be effectively disabled. I think if people enable keepalives, they will want to have connection close when acks aren't received back.

src/gun_http2.erl
```diff
@@ -852,9 +857,15 @@ close_stream(State, #stream{ref=StreamRef, reply_to=ReplyTo}, Reason) ->
 	ReplyTo ! {gun_error, self(), stream_ref(State, StreamRef), Reason},
 	ok.
 
-keepalive(#http2_state{socket=Socket, transport=Transport}, _, EvHandlerState) ->
+keepalive(State=#http2_state{pings_sent=PingsSent, opts=Opts}, _, EvHandlerState)
+		when PingsSent > map_get(Opts, keepalive_tolerance) ->
```
essen, line 861: Enabling by default would mean we can't use (only) `map_get` though.

essen (commented):
Please continue.

src/gun_http2.erl
zuiderkwast: Yes, makes more sense to accept a delay but eventually all pings need to be acked.

essen: That's not what I mean. If you send 3 pings and then receive a ping_ack, pings_sent will go from 3 to 0, and then when you send more pings and receive ping_acks you're out of sync with reality.

zuiderkwast: Ah, right. That can make us keep a connection with even if it has a very long network delay.

My interpretation of the requirement was that we need to receive one ack for every N pings sent, so the increment and reset achieves this. We could achieve this with a different payload for each ping and match each ack to its ping. But I don't think the extra bookkeeping is worth it and furthermore there is no reason to accept data loss.

doc/src/manual/gun.asciidoc
zuiderkwast: OK, what number do you think we should use as a default? 3?

src/gun_http2.erl
essen: I am not saying we need to match the ping content (that's actually already done in the protocol code IIRC), just that the count needs to be accurate. We send one ping, we +1. We receive a ping_ack, we -1.

doc/src/manual/gun.asciidoc
essen: On second thought let's not set a value by default, instead let's mention this option inside the `keepalive` option description.

src/gun_http2.erl
zuiderkwast: I understand that and I'm saying that if we do that, it implies that we require an ack for each ping, not just one for every N pings. And that's fine.

essen: Right. The relevant spec is at https://datatracker.ietf.org/doc/html/rfc7540#section-6.7

```
   Receivers of a PING frame that does not include an ACK flag MUST send
   a PING frame with the ACK flag set in response, with an identical
   payload.  PING responses SHOULD be given higher priority than any
   other frame.
```

So you will get an ack for every ping.

doc/src/manual/gun.asciidoc
zuiderkwast: Done.

src/gun_http2.erl
zuiderkwast: Not enabling by default.

essen:
Please try to avoid these merge commits [Merge remote-tracking branch 'origin/master' into keepalive_tolerance] they make things much harder for me to merge afterwards. `git pull --rebase origin master` is the way to go (where `origin` is this repository, on my machine it's named `upstream`).

zuiderkwast:
OK, I'll revert the merge and do a rebase instead. (Some projects prefer merge instead of force-push since it makes it easier to see what's already reviewed and what's new.)

essen:
No problem, thanks.

essen:
Merged with many tweaks, even went back on some things I said before. Thanks!

essen:
Hey, there's some issues in the tests for this on macOS and Windows. Can you please take a look? If you open a PR it'll run the tests again.

zuiderkwast:
It looks like a timing issue. If something is lagging by 50ms, the `gun_down` comes too early.

The simplest solution would be to increase the times, making the test take longer to run. Perhaps by a factor 10. Do you have a better idea?

This is from one of the latest mac builds on master:

image

essen:
If it's a timing issue then yes increasing would be fine for the time being. But it's a bit surprising because macOS tends to be the fastest to run tests and have the least amount of issues from it.

zuiderkwast:
OK, here's a boring PR increasing the timeouts 10 times: #309. Let's see if all builds pass.

Are you in control of the mac machines? Maybe they were temporarily heavily loaded or something....

essen:
It's possible for Windows since it's a VM on my (beefy) gaming PC. But macOS is CI-only. I'm in control of all machines so if it's not something trivial I'll be able to take a closer look next week.

PR #295, opened 2022-05-27
Implemented 'start_pool_if_missing' logic

umbec:
Hi @essen!

I have this patch that expands the `start_pool_if_missing` logic, allocating a new pool on demand.

If a pool for a certain authority already exists, the checkout process succeed, and the request goes on as always.
In the other case, a new pool is allocated with the opts() passed to the `start_pool_if_missing` flag.

This code has been working on my services for a while, and without the presumption of being perfect or to address all the open points in the pool management, it can brings an useful improvement to the current implementation.

If you have any objection or suggestion to the implementation, let me know.

### DIFF
--- a/src/gun_pool.erl
+++ b/src/gun_pool.erl
@@ -125,7 +125,7 @@
 	checkout_call_timeout => timeout(),
 	checkout_retry => [pos_integer()],
 	scope => any(),
-	start_pool_if_missing => boolean()
+	start_pool_if_missing => false | opts()
 }.
 -export_type([req_opts/0]).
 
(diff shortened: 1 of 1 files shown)


essen:
Thanks! I will have to review and that will take a while as I'm in the middle of a very heavy set of changes. But the initial feeling is good.

essen:
OK I took a lot of time...

I will not be merging this because this is not what the feature should be. There's two types of non-static pools: one is a temporary pool where the pool stops automatically when nothing happens, the other is a dynamic pool where the size of the pool varies depending on what's going on. These are the ones I envision for `start_pool_if_missing`.

I can't see the option fitting for static pools, because static pools stay up forever. They never downsize to 0, they keep the connection(s) to the server open forever.

I can understand wanting to start static pools in a more dynamic manner, but this doesn't require modifying `gun_pool`: one can just use `gun_pool:info` to figure out if the pool already exists, and start it.

Closing, thanks!!

PR #296, opened 2022-06-06
Allows SSE with charset in content type

tank-bohr:
When content type is something like `text/event-stream;charset=UTF-8` gun didn't send sse-related events.


tank-bohr:
Looks like a duplicate of #271

PR #298, opened 2022-08-17
deprecated erlang:get_stacktrace/0 function

velimir:
This to allow one to build the lib on erlang 24 and higher


velimir:
good day @essen ! Is there any chance this would get some attention from your side in the nearest future?

essen:
I am not working on Gun 1.x anymore, I recommend moving to Gun 2.0 (even RC) as that one is maintained.

PR #299, opened 2022-08-25
Make Protocol:init/4 return an ok-tuple

zuiderkwast:
This is a preparation for allowing init/4 to return an ok or an
error tuple.

This part was done as part preparation for handling send errors as discussed here:
https://github.com/ninenines/gun/pull/243#discussion_r824856311


essen:
Woot! I'm back to work on Tuesday.

essen:
Merged, thanks!

PR #301, opened 2022-09-27
Make many gun_http2 functions return state or error tuple

zuiderkwast:
The following functions used to return a state, but now return {state, State} or {error, Reason}:

* frame/5
* update_window/1,2
* maybe_ack_or_notify/2
* reset_stream/3
* push_promise_frame/7
* goaway/2
* maybe_send_data/6
* send_data/4
* send_data/6
* send_data_frame/4

----

This is incremental refactoring preparing for #243

### DIFF
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -450,6 +450,7 @@ data_frame1(State0, StreamID, IsFin, Data, EvHandler, EvHandlerState0,
 			{{state, maybe_delete_stream(State, StreamID, remote, IsFin)},
 				EvHandlerState};
 		Error={error, _} ->
+			%% @todo Delete stream and return new state and error commands.
 			{Error, EvHandlerState}
 	end.
 
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -243,7 +243,7 @@ parse(Data, State0=#http2_state{status=Status, http2_machine=HTTP2Machine, strea
 				{state, State} ->
 					parse(Rest, State, CookieStore0, EvHandler, EvHandlerState0);
 				Error={error, _} ->
-					Error
+					{Error, CookieStore0, EvHandlerState0}
 			end;
 		Error = {connection_error, _, _} ->
 			{connection_error(State0, Error), CookieStore0, EvHandlerState0};
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -360,9 +360,9 @@ maybe_ack_or_notify(State=#http2_state{reply_to=ReplyTo, socket=Socket,
 data_frame(State0, StreamID, IsFin, Data, CookieStore0, EvHandler, EvHandlerState0) ->
 	case get_stream_by_id(State0, StreamID) of
 		Stream=#stream{tunnel=undefined} ->
-			{StateOrError, EvHandlerState} = data_frame1(State0,
+			{Commands, EvHandlerState} = data_frame1(State0,
 				StreamID, IsFin, Data, EvHandler, EvHandlerState0, Stream),
-			{StateOrError, CookieStore0, EvHandlerState};
+			{Commands, CookieStore0, EvHandlerState};
 		Stream=#stream{tunnel=#tunnel{protocol=Proto, protocol_state=ProtoState0}} ->
 %			%% @todo What about IsFin?
 			{Commands, CookieStore, EvHandlerState1} = Proto:handle(Data,
@@ -445,13 +445,13 @@ data_frame1(State0, StreamID, IsFin, Data, EvHandler, EvHandlerState0,
 					{update_window(State1), EvHandlerState1}
 			end
 	end,
-	case StateOrError of
-		{state, State} ->
-			{{state, maybe_delete_stream(State, StreamID, remote, IsFin)},
-				EvHandlerState};
+	{State, Errors} = case StateOrError of
+		{state, State2} ->
+			{State2, []};
 		Error={error, _} ->
-			{Error, EvHandlerState}
-	end.
+			{State1, [Error]}
+	end,
+	{[{state, maybe_delete_stream(State, StreamID, remote, IsFin)} | Errors], EvHandlerState}.
 
 headers_frame(State0=#http2_state{opts=Opts},
 		StreamID, IsFin, Headers, #{status := Status}, _BodyLen,
--- a/src/gun_http2.erl
+++ b/src/gun_http2.erl
@@ -203,7 +203,7 @@ parse(Data, State0=#http2_state{status=preface, http2_machine=HTTP2Machine},
 			case frame(State0#http2_state{status=connected}, Frame, CookieStore0, EvHandler, EvHandlerState0) of
 				{Error={error, _}, CookieStore, EvHandlerState} ->
 					{Error, CookieStore, EvHandlerState};
-				{State, CookieStore, EvHandlerState} ->
+				{{state, State}, CookieStore, EvHandlerState} ->
 					parse(Rest, State, CookieStore, EvHandler, EvHandlerState)
 			end;
 		more ->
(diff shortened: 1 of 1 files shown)


zuiderkwast:
Good morning Loïc. Are you busy?

I believe when this is merged, the remaining "handle send errors" will be a very small change.

essen:
Sounds good. Yes if I don't get to this today it will be next Monday.

zuiderkwast:
OK, great. The change is large but trivial, so I hope it is easy to review.

Then, I can help finishing the remaining 2.0 projects, if you know what you want with them. :-)

src/gun_http2.erl
```diff
@@ -352,14 +357,14 @@ maybe_ack_or_notify(State=#http2_state{reply_to=ReplyTo, socket=Socket,
...
-	State.
+	{state, State}.
 
 data_frame(State0, StreamID, IsFin, Data, CookieStore0, EvHandler, EvHandlerState0) ->
 	case get_stream_by_id(State0, StreamID) of
 		Stream=#stream{tunnel=undefined} ->
 			{State, EvHandlerState} = data_frame1(State0,
 				StreamID, IsFin, Data, EvHandler, EvHandlerState0, Stream),
-			{State, CookieStore0, EvHandlerState};
+			{{state, State}, CookieStore0, EvHandlerState};
```
essen, line 367: That can't be right, `data_frame1` can return an error in the code you add in this PR.

src/gun_http2.erl
```diff
@@ -290,41 +294,42 @@ frame(State=#http2_state{http2_machine=HTTP2Machine0}, Frame, CookieStore, EvHan
...
-			case StateRet of
-				{state, State1} -> {State1, CookieStoreRet, EvHandlerStateRet};
-				Error -> {Error, CookieStoreRet, EvHandlerStateRet}
-			end;
+			data_frame(State#http2_state{http2_machine=HTTP2Machine},
+				StreamID, IsFin, Data, CookieStore, EvHandler, EvHandlerState);
 		{ok, {headers, StreamID, IsFin, Headers, PseudoHeaders, BodyLen}, HTTP2Machine} ->
-			headers_frame(State#http2_state{http2_machine=HTTP2Machine},
+			{StateRet, CookieStoreRet, EvHandlerStateRet} = headers_frame(
+				State#http2_state{http2_machine=HTTP2Machine},
```
essen, line 301: This is a general comment regarding both data_frame, headers_frame and the other *_frame functions: they should follow the same return value format. So if data_frame changes to a tuple return value, they all should.

src/gun_http2.erl
```diff
@@ -438,7 +447,12 @@ data_frame1(State0, StreamID, IsFin, Data, EvHandler, EvHandlerState0,
 					{update_window(State1), EvHandlerState1}
 			end
 	end,
-	{maybe_delete_stream(State, StreamID, remote, IsFin), EvHandlerState}.
+	case StateOrError of
+		{state, State} ->
+			{maybe_delete_stream(State, StreamID, remote, IsFin), EvHandlerState};
+		Error={error, _} ->
+			{Error, EvHandlerState}
+	end.
```
essen, line 455: We should probably maybe_delete_stream regardless of the error just so we keep the state consistent.

essen: Of course we don't yet have errors here but we will in update_window which may happen even after the stream has terminated and so we should update the state.

src/gun_http2.erl
```diff
@@ -1074,63 +1087,76 @@ maybe_send_data(State=#http2_state{http2_machine=HTTP2Machine0}, StreamID, IsFin
...
+					#stream{ref=StreamRef, reply_to=ReplyTo} = get_stream_by_id(State, StreamID),
+					RequestEndEvent = #{
+						stream_ref => stream_ref(State, StreamRef),
+						reply_to => ReplyTo
+					},
+					EvHandler:request_end(RequestEndEvent, EvHandlerState0)
+			end,
+			{{state, maybe_delete_stream(State, StreamID, local, IsFin)}, EvHandlerState};
+		Error={error, _Reason} ->
+			{Error, EvHandlerState0}
```
essen, line 1122: As a counter-example to the previous comment, here if we fail to send the final data frame it makes sense to not update the state. So it's OK to keep the maybe_delete_stream for only successful sends as you did.

essen: No changes needed here.

essen:
@zuiderkwast Reviewed. There's at least one error and a couple consistency problems that should easily be resolved.

src/gun_http2.erl
zuiderkwast: To be clear, no changes needed here (`data_frame1`) since it's the same situation as in `send_data` (the other comment), right?

essen: No this one is needed because it's NOT the same situation: the stream has really ended, all data for the stream is sent/received, but a window_update failed after. In send_data it is the sending of the stream's data that failed so the stream didn't end and if there's a problem we probably want to know about that stream.

src/gun_http2.erl
zuiderkwast: Updating the following functions return state tuple or error: `headers_frame`, `trailers_frame`, `rst_stream_frame`, `headers_frame_inform`, `headers_frame_connect`, `headers_frame_connect_websocket`, `headers_frame_response`, `ignored_frame`.

src/gun_http2.erl
zuiderkwast: `data_frame1` updated to return state tuple.

src/gun_http2.erl
zuiderkwast: OK, thanks for explaining.

If we return `{state, maybe_delete_stream(State0, ...)}` even in the error case, what to do with the error? Ignore it?

essen: Would it be a problem to return `[{state, maybe...(State)}, {error, Reason}]`? If it's a problem, please just add a `@todo` comment for me to fix later.

zuiderkwast: It seems to work. Thx.

zuiderkwast: Hrm, it doesn't work, because of code matching on state or error, like below. I'm adding a todo instead.

```Erlang
parse(Data, State0=#http2_state{status=Status, http2_machine=HTTP2Machine, streams=Streams},
		CookieStore0, EvHandler, EvHandlerState0) ->
	MaxFrameSize = cow_http2_machine:get_local_setting(max_frame_size, HTTP2Machine),
	case cow_http2:parse(Data, MaxFrameSize) of
		{ok, Frame, Rest} ->
			case frame(State0, Frame, CookieStore0, EvHandler, EvHandlerState0) of
				{Error={error, _}, CookieStore, EvHandlerState} ->
					{Error, CookieStore, EvHandlerState};
				{{state, State}, CookieStore, EvHandlerState} ->
					parse(Rest, State, CookieStore, EvHandler, EvHandlerState)
			end;
```

zuiderkwast: I can normalize the result of frame/1 using a function like `commands/N` found in other modules. Is this what you have in mind?

```Erlang

			{Commands, CookieStore, EvHandlerState} = frame(State0, Frame,
				CookieStore0, EvHandler, EvHandlerState0),
			case commands(Commands, State0) of
				{state, State} ->
					parse(Rest, State, CookieStore, EvHandler, EvHandlerState)
				ResCommands ->
					{ResCommands, CookieStore, EvHandlerState};
			end;
```

essen: It might end up necessary yes. If you do that please keep it as a separate commit.

zuiderkwast: I can save it for another PR, if you think this one can be merged as is.

zuiderkwast: I've pushed this change as a commit in separate branch (on top of this PR): https://github.com/Nordix/gun/commit/9340a1cbc28eed7c062daa72f7a0f9e49414ddaa

If you like it, I can add id here in this PR. (I'm not sure if I like it or not.)

essen: Hm that's not what I had in mind. Let's just keep the todo for now. Review was good this time so I'll proceed with merge now.

essen:
Merged, thanks! I've also fixed the TLS tests that fail. Dialyzer will fail until the other PR is merged.

Issue #303, opened 2022-11-09
gun:connect expects proxy server to reply with HTTP/1.1, some servers respond with HTTP/1.0


JesseStimpson:
Hello,

I'm attempting to connect to an HTTP proxy that is running [goproxy](https://github.com/stripe/goproxy) (the project is from Stripe, but I am not affiliated with them) and encountering a failure in HTTP version incompatibility. Here is my client code

```
fun() ->
    application:ensure_all_started(gun),
    {ok, ConnPid} = gun:open("my-https-proxy.example.com", 443,
        #{protocols => [http], transport => tls, tls_opts => [{verify, verify_none}]}),
    {ok, http} = gun:await_up(ConnPid),
    ConnectRef = gun:connect(ConnPid,
        #{host => "www.google.com",
          port => 443,
          protocols => [http],
          transport => tls,
          tls_opts => [{verify, verify_none}]
          }),
    {response, fin, 200, _} = gun:await(ConnPid, ConnectRef),
    GetRef = gun:get(ConnPid, "/", #{}, #{tunnel => ConnectRef}),
    Result = gun:await(ConnPid, GetRef),
    gun:close(Result),
    Result
end().
```

This results in the following error
```
=CRASH REPORT==== 9-Nov-2022::21:42:16.365258 ===
  crasher:
    initial call: gun:init/1
    pid: <0.196.0>
    registered_name: []
    exception error: no function clause matching
                     gun_http:handle_connect(<<>>,
                                             {http_state,
                                              {sslsocket,
                                               {gen_tcp,#Port<0.16>,
                                                tls_connection,undefined},
                                               [<0.199.0>,<0.198.0>]},
                                              gun_tls,#{},'HTTP/1.1',
                                              keepalive,<<>>,undefined,
                                              [{stream,
                                                {connect,
                                                 #Ref<0.3830098819.398983169.252231>,
                                                 #{host => "www.google.com",
                                                   port => 443,
                                                   protocols => [http],
                                                   tls_opts =>
                                                    [{verify,verify_none}],
                                                   transport => tls}},
                                                <0.187.0>,infinity,
                                                <<"CONNECT">>,
                                                ["www.google.com",58,
                                                 <<"443">>],
                                                <<>>,true,undefined}],
                                              head,
                                              {0,0},
                                              head},
                                             undefined,gun_default_event_h,
                                             undefined,'HTTP/1.0',200,[]) (/home/jstimpson/gun/src/gun_http.erl, line 311)
[...]

PR #304, opened 2022-11-10
Add allow_http10_connect in http_opts

JesseStimpson:
When true, this option permits out-of-spec proxy servers that respond with HTTP/1.0 on the response to the CONNECT. Default behaviour is to only allow HTTP/1.1.

This PR is submitted to address #303 .

These changes pass my local testing, which includes:
1. A successful tunnel through goproxy with `allow_http10_connect => true`, which responds with HTTP/1.0 on the CONNECT response. ~~[goproxy issue](https://github.com/goproxy/goproxy/issues/38)~~[1]
2. An new error response[0] with `allow_http10_connect => false` when HTTP/1.0 is provided.
3. make eunit
4. make ct

I've not attempted to add any test cases or documentation yet. The test cases for tunneling are quite comprehensive, so I would need some guidance on the best way to approach this.

[0] New error response:
```
** exception error: no match of right hand side value {error,{stream_error,{badstate,"CONNECT cannot be used over an HTTP/1.0 connection by default. See http_opts `allow_http10_connect`."}}}
```

[1] Edit: I opened the referenced goproxy issue under the wrong project by mistake, so please ignore this particular link for now. If I am able to open the correct issue, I will edit this message later on.

### DIFF
--- a/doc/src/manual/gun.asciidoc
+++ b/doc/src/manual/gun.asciidoc
@@ -153,7 +153,8 @@ http_opts() :: #{
     flow                        => pos_integer(),
     keepalive                   => timeout(),
     transform_header_name       => fun((binary()) -> binary()),
-    version                     => 'HTTP/1.1' | 'HTTP/1.0'
+    version                     => 'HTTP/1.1' | 'HTTP/1.0',
+    allow_http10_connect        => boolean()
 }
 ----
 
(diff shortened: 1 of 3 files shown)


essen:
After thinking some time about this I think I will enable HTTP/1.0 responses without the flag. While RFC7231 doesn't say anything about the version in responses to CONNECT requests, RFC7230 establishes version rules and I don't think they prevent the use of HTTP/1.0. Also in practice because of how old CONNECT is I don't think there would be any HTTP/1.0 proxies that response positively and yet do not implement it properly (not that there is much it can fail on). I will do the change (just removing the version argument) and add a test.

essen:
Pushed, please confirm it works for you (but probably does considering the change). Thanks!

JesseStimpson:
I will check our usage, but based on the commit I'm sure it will work as expected. Thanks for the fix!

PS: It looks like #277 can be closed as well.

PR #309, opened 2023-01-16
Increase times by factor 10 to fix timing issue

zuiderkwast:
Attempt to fix unstable test case `keepalive_tolerance_ping_ack_timeout`.

Makes the test less sensitive to slow test runners.


essen:
Good enough for the time being. Thanks!

zuiderkwast:
OK, good. It didn't make all builds green, but I suspect there is a similar timing issue (or some other issue) in the shutdown suite.

PR #319, opened 2023-09-25
HTTP/3

essen:
Companion PR to https://github.com/ninenines/cowboy/pull/1599


essen:
The failures on macOS are due to GH Actions macOS runner having trouble with DNS. I suggest leaving the failures for now as there's no release immediately planned.

essen:
Merged!!

PR #336, opened 2024-05-14
Include event handler state in gun:info/1 result

zuiderkwast:
Adds event_handler and event_handler_state in the map returned by gun:info/1.


essen:
Wondering what you use these for? If it's only for debugging, `sys:get_state` will get them just fine.

essen:
Right, makes sense.

essen:
Merged, thanks!

PR #340, opened 2024-07-09
do not ignore data received immediately after switching to raw

chilikk:
When an HTTP connection is upgraded to raw it may happen that the body/the data belonging to the raw connection arrives in the same message as HTTP response and headers from the TCP socket. This change fixes the bug where such data has been dropped. With this fix the data following the HTTP response is forwarded to the `gun_raw` handler.


chilikk:
Changed according to your suggestion. I'm not exactly sure how to handle all `switch_protocol`: the same handling could possibly apply to `gun_http:handle_connect` case, but I'm unsure what would be the scenario and how to test it, so I didn't change. Also, in `gun_tunnel` I don't handle any other case than `<<>>`, because my understanding is that `gun_tunnel` is only used from HTTP/2 and this situation would not be possible there due to the framing mechanism.

essen:
Merged, thanks!

PR #343, opened 2024-12-13
Add gun:ping/2,3 for user-initiated ping for HTTP/2

zuiderkwast:
This feature allows a user to explicitly send a ping frame and receive a ping ack. The use case is to test the latency and health of existing long-lived API connections using some admin tool.

A request, such as OPTIONS or HEAD, can also be used for this purpose, but it may be handled differently by middleware.

This PR implements it only for HTTP/2, which is the use case we have. For other protocols, an error tuple is sent to the reply-to process.

This use case is mentioned in RFC 7540, section 6.7:

    The PING frame (type=0x6) is a mechanism for measuring a minimal
    round-trip time from the sender, as well as determining whether an
    idle connection is still functional.

Section 8.1.4:

    The PING frame allows a client to safely test whether a connection is
    still active without sending a request.

### DIFF
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -60,6 +60,10 @@
 %% Streaming data.
 -export([data/4]).
 
+%% User pings.
+-export([ping/1]).
+-export([ping/2]).
+
 %% Tunneling.
 -export([connect/2]).
 -export([connect/3]).
@@ -721,6 +725,20 @@ data(ServerPid, StreamRef, IsFin, Data) ->
 			gen_statem:cast(ServerPid, {data, self(), StreamRef, IsFin, Data})
 	end.
 
+%% User pings.
+
+-spec ping(pid()) -> stream_ref().
+ping(ServerPid) ->
+	ping(ServerPid, #{}).
+
+-spec ping(pid(), req_opts()) -> stream_ref().
+ping(ServerPid, ReqOpts) ->
+	Tunnel = get_tunnel(ReqOpts),
+	StreamRef = make_stream_ref(Tunnel),
+	ReplyTo = maps:get(reply_to, ReqOpts, self()),
+	gen_statem:cast(ServerPid, {ping, ReplyTo, StreamRef}),
+	StreamRef.
+
 %% Tunneling.
 
 -spec connect(pid(), connect_destination()) -> stream_ref().
@@ -782,6 +800,8 @@ await(ServerPid, StreamRef, Timeout, MRef) ->
 			{up, Protocol};
 		{gun_notify, ServerPid, Type, Info} ->
 			{notify, Type, Info};
+		{gun_ping_ack, ServerPid, StreamRef} ->
+			ping_ack;
 		{gun_error, ServerPid, StreamRef, Reason} ->
 			{error, {stream_error, Reason}};
 		{gun_error, ServerPid, Reason} ->
@@ -1367,6 +1387,15 @@ connected_ws_only(Type, Event, State) ->
 %%
 %% @todo It might be better, internally, to pass around a URIMap
 %% containing the target URI, instead of separate Host/Port/PathWithQs.
+connected(cast, {ping, ReplyTo, StreamRef},
+		State=#state{protocol=Protocol, protocol_state=ProtoState}) ->
+	case erlang:function_exported(Protocol, ping, 3) of
+		true ->
+			Commands = Protocol:ping(ProtoState, dereference_stream_ref(StreamRef, State), ReplyTo),
+			commands(Commands, State);
+		false ->
+			ReplyTo ! {gun_error, self(), StreamRef, not_supported_for_protocol}
+	end;
 connected(cast, {headers, ReplyTo, StreamRef, Method, Path, Headers, InitialFlow},
 		State=#state{origin_host=Host, origin_port=Port,
 			protocol=Protocol, protocol_state=ProtoState, cookie_store=CookieStore0,
--- a/src/gun.erl
+++ b/src/gun.erl
@@ -800,8 +800,6 @@ await(ServerPid, StreamRef, Timeout, MRef) ->
 			{up, Protocol};
 		{gun_notify, ServerPid, Type, Info} ->
 			{notify, Type, Info};
-		{gun_ping_ack, ServerPid, StreamRef} ->
-			ping_ack;
 		{gun_error, ServerPid, StreamRef, Reason} ->
 			{error, {stream_error, Reason}};
 		{gun_error, ServerPid, Reason} ->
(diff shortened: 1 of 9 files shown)


src/gun.erl
```diff
@@ -1367,6 +1387,15 @@ connected_ws_only(Type, Event, State) ->
...
 %% containing the target URI, instead of separate Host/Port/PathWithQs.
+connected(cast, {ping, ReplyTo, StreamRef},
+		State=#state{protocol=Protocol, protocol_state=ProtoState}) ->
+	case erlang:function_exported(Protocol, ping, 3) of
+		true ->
+			Commands = Protocol:ping(ProtoState, dereference_stream_ref(StreamRef, State), ReplyTo),
+			commands(Commands, State);
+		false ->
+			ReplyTo ! {gun_error, self(), StreamRef, not_supported_for_protocol}
+	end;
```
essen, line 1398: Better not have optional callbacks for protocols. Rather make them return an error, and check the return value.

src/gun_http2.erl
```diff
@@ -934,6 +957,20 @@ keepalive(State=#http2_state{socket=Socket, transport=Transport, pings_unack=Pin
 			{Error, EvHandlerState}
 	end.
 
+ping(State=#http2_state{socket=Socket, transport=Transport, user_pings=UserPings}, StreamRef, ReplyTo) ->
+	%% Use non-zero 64-bit payload for user pings. 0 is reserved for keepalive.
+	Payload = case erlang:monotonic_time(microsecond) band 16#ffffffffffffffff of
+		0 -> 1;
+		Payload0 -> Payload0
+	end,
```
essen, line 965: Just have a counter that starts at 1 (or 1001, or other...) and increment every time, it'll save us trouble.

essen (changes requested):

src/gun_http2.erl
zuiderkwast: I guess I wanted avoid wasting memory to store a counter for a feature that almost nobody will use.

What trouble do you anticipate?

essen: Future changes may use more than 0 for keepalive and we don't want conflicts to happen. The counter can be stored in `user_pings` when user pings are used, and `user_pings` can be set to `false` when they are not used.

src/gun.erl
zuiderkwast: We have a `has_keepalive() -> boolean()` just to check if a protocol supports keepalive. Shall we follow this style and add `has_ping() -> boolean()`?

essen: I don't think they are different, the only difference will be whether they have been implemented, and that can be signified via the return value.

We need `has_keepalive` because we must start a timer. We don't need that for user pings, we can just call the function directly and propagate the error if any.

I think we should make a difference between protocols that will implement it in the future (http3 and ws are likely candidates, would be `not_implemented`) and protocols that won't (http, raw, socks... would be `unsupported_by_protocol` or something less verbose).

src/gun.erl
```diff
@@ -782,6 +800,8 @@ await(ServerPid, StreamRef, Timeout, MRef) ->
 			{up, Protocol};
 		{gun_notify, ServerPid, Type, Info} ->
 			{notify, Type, Info};
+		{gun_ping_ack, ServerPid, StreamRef} ->
+			ping_ack;
```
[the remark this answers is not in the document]
essen: I had not seen that PingRef. The problem with doing it that way is that it circumvents a stream-specific concept (StreamRef) and hijacks it for a connection-specific thing (pings). So I do not like that approach.

I also wonder if gun:await would be useful for this outside of tests. I would say it's OK to use `gun_notify`. Worst case, the message is documented, so a selective receive lets you receive exactly the message you are looking for if necessary. In the case of pings I think that would be `{gun_notify, ServerPid, ping_ack, PingRef}`.

src/gun_http2.erl
essen: `erlang:unique_integer([monotonic, positive])` is probably good enough.

zuiderkwast: Sure, that's fine too. Which do you prefer: unique integer or counter?

src/gun.erl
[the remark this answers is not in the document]
zuiderkwast: Yeah, selective receive is probably fine.

Are you OK with hijacking the stream ref mechanism for the rest of this implementation?

I was thinking like this: A ping is similar to a request, so the stream ref is a nice abstraction for a user. In HTTP/2 it's a connection level message but it could have been a request. It's a protocol detail.

src/gun_http2.erl
essen: The unique integer function is a kind of counter. It will work better than keeping a counter ourselves.

``` erlang
1> erlang:unique_integer([monotonic, positive]).
1
2> erlang:unique_integer([monotonic, positive]).
2
3> erlang:unique_integer([monotonic, positive]).
3
4> erlang:unique_integer([monotonic, positive]).
4
```

src/gun.erl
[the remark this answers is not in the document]
essen: In HTTP/3 it is a QUIC frame, so it sits even below the protocol. I don't think it should be described or used like a StreamRef. It's a PingRef. No problem having two references used for different purposes. There's no hijacking that I could see outside of the message so if `gun_notify` is used I'm not sure I see a problem.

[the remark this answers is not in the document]
zuiderkwast: OK, makes sense. I hijacked the stream error format too for the `{gun_error, ServerPid, StreamRef, not_supported_by_protocol}` error, so `gun:await` for the PingRef would catch it. I'll remove this special case and treat it as a connection error so it causes a disconnect. Less special cases.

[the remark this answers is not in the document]
zuiderkwast: I used `dereference_stream_ref` on the PingRef too. I guess it works for tunneling.

src/gun.erl
zuiderkwast: They can just return `{error, not_implemented}` and `{error, unsupported_by_protocol}` and we don't even need to check the return value here. Then they'll be handled like any connection error and it leads to a disconnect. Does this seem OK?

Alternatively, I can wrap it in another tuple like `{ok, Commands} | {error, UnsupportedReason}`. I don't see the need for this now but I can do it if you prefer it.

src/gun_http2.erl
zuiderkwast: Sure. Unique integer does some inter-process sync so it may be slower but that's not a concern here. This is not a hot code patch.

src/gun.erl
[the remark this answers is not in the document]
essen: I'd say for tunneling you would have a StreamRef refering to the tunneled connection, but the PingRef would be kept separate, since they're not the same thing.

I don't know whether using ping on a protocol that doesn't support it should cause a disconnect (it can if that's all we have today), but it certainly should cause a connection error forwarded to the user.

src/gun.erl
essen: The error reasons seem OK to me.

src/gun_http2.erl
essen: Same as monotonic_time in that regard, but better control over the values it returns.

Make it send payloads starting at 10000 (`10000 + erlang:unique_integer([monotonic, positive])`) so there's plenty of space left if we need other values in the future. Not sure we need to `band 16#ffffffffffffffff`, I don't know how long that would take for unique_integer to reach that value, but if we do the wrapped value should be > 10000.

essen:
I have opened https://github.com/ninenines/gun/pull/349 with a tunnel implementation. I will look into maybe implementing this for Websocket as well as that has been requested before.

essen:
The other PR was merged. Thanks!!

PR #344, opened 2025-01-02
HTTP/2: Fix tunneled streams bugs

essen:
* Tunneled streams can now close the stream.

* Data received on tunneled streams now result in WINDOW_UPDATE frames being sent if necessary, and flow control is handled.

This was detected as part of [writing a new Cowboy test suite for benchmarking Websocket](https://github.com/ninenines/cowboy/pull/1658), but should help other uses too.


essen:
Merged.

PR #349, opened 2025-03-26
User ping

essen:
Full implementation of https://github.com/ninenines/gun/pull/343


essen:
OK for Websocket we can already do this by sending ping frames and setting `silence_pings` to `false` so I guess there's no need to duplicate the functionality for now. We can unify all this when implementing e.g. HTTP/3 user pings.

essen:
Merged!!

PR #351, opened 2025-04-12
Fix testing with rebar

lemenkov:
We're using Rebar3 for everything from building to testing - fetching deps, preparing pre-built steps, building, and testing. Unfortunately your rebar.config file lacks ct_opts parameters so it fails the tests. This patch fixes that (and doesn't change anything for erlang.mk-based workflows).


essen:
Please use an override instead.

essen:
The rebar.config file is generated so this would be lost on next generate.

PR #359, opened 2025-11-16
Add gun:trailers/3 for HTTP/1.1 and HTTP/2

### LINKED ISSUE #356: Support for HTTP/1.1 chunked trailers
Hey hey! :wave:

Poking around the `gun` functions I don't think it's currently possible to send HTTP/1.1 trailing headers ([RFC7230 s. 4.1.2](https://www.rfc-editor.org/rfc/rfc7230#section-4.1.2)) when using chunked transfer-encoding.

Like one of the examples in the RFC, I'd like to use a trailing header for message integrity while uploading a file to S3 ([docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#trailing-checksums)). So I would read a chunks of a file from disk and stream on the connection while building a CRC32 or MD5, and then send that checksum as a trailer.

For other prior art I know that you can send trailers with [`Mint.HTTP.stream_request_body/3`](https://hexdocs.pm/mint/Mint.HTTP.html#stream_request_body/3) which is conceptually very similar to `gun:data/3`. Maybe a nice way to handle this in the `gun:data/3` interface could be to allow sending trailers with `fin` like Mint's `{:eof, headers()}` form.

```erlang
-spec data(ConnPid, StreamRef, IsFin, Data) -> ok when
    ConnPid :: pid(),
    StreamRef :: gun:stream_ref(),
    IsFin :: fin | {fin, gun:req_headers()} | nofin,
    Data :: iodata().
```

Then you could stream the final chunk of the request body and the trailers in the same call to `data/3`.

What do you think - would you be interested in a PR for this?

the-mikedavis:
This is a work-in-progress for adding `gun:trailers/3`. So far I adapted what `data/7` is doing in HTTP/1 and HTTP/2 - I still need to look at adding test cases.

Closes https://github.com/ninenines/gun/issues/356

### DIFF
--- a/doc/src/guide/protocols.asciidoc
+++ b/doc/src/guide/protocols.asciidoc
@@ -111,6 +111,7 @@ current protocol.
 | put        | yes      | yes    | no
 | request    | yes      | yes    | no
 | data       | yes      | yes    | no
+| trailers   | yes      | yes    | no
 | await      | yes      | yes    | no
 | await_body | yes      | yes    | no
 | flush      | yes      | yes    | no
--- a/doc/src/manual/gun.asciidoc
+++ b/doc/src/manual/gun.asciidoc
@@ -34,6 +34,7 @@ Requests:
 * link:man:gun:headers(3)[gun:headers(3)] - Initiate the given request
 * link:man:gun:request(3)[gun:request(3)] - Perform the given request
 * link:man:gun:data(3)[gun:data(3)] - Stream the body of a request
+* link:man:gun:trailers(3)[gun:trailers(3)] - Send the request trailers
 
 Proxies:
 
--- a/doc/src/manual/gun.data.asciidoc
+++ b/doc/src/manual/gun.data.asciidoc
@@ -73,3 +73,4 @@ link:man:gun:patch(3)[gun:patch(3)],
 link:man:gun:post(3)[gun:post(3)],
 link:man:gun:put(3)[gun:put(3)],
 link:man:gun:headers(3)[gun:headers(3)]
+link:man:gun:trailers(3)[gun:trailers(3)]
--- a/doc/src/manual/gun.trailers.asciidoc
+++ b/doc/src/manual/gun.trailers.asciidoc
@@ -1,0 +1,76 @@
+= gun:trailers(3)
+
+== Name
+
+gun:trailers - Send the request trailers
+
+== Description
+
+[source,erlang]
+----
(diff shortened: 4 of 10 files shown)


essen:
This will need to be made to work with HTTP/3 as well. It likely already works if you do the same as HTTP/2. Tests will confirm. Worst case, using trailers with HTTP/3 should trigger an error (there's plenty of things not implemented with HTTP/3 so if tests show it doesn't work it's fine to error out for now).

the-mikedavis:
For HTTP/3, a case that errors is probably the easiest (for me at least :sweat_smile:) for now. It looks like we need to access/update the `#http3_state.http3_machine` for the QPACK encoding. (I'm pretty familiar with HTTP/2 but not HTTP/3 - maybe this is easier than I am thinking.) `gun_http3:data/7` uses `cow_http3:data/1` but I think we want to use `cow_http3:headers/1` on the headers block from `cow_http3_machine:prepare_trailers/3` instead.

essen:
Right if it's not the same as HTTP/2 let's error out for now in gun_http3.

the-mikedavis:
Sorry for the really long delay :sweat_smile:. I ran into (what I think is?) a bug in cowlib for HTTP/2. I sent a PR over there with the fix, https://github.com/ninenines/cowlib/pull/149.

the-mikedavis:
Ultimately I ended up not using this change since I'm targetting S3 and it has a weird `content-encoding: aws-chunked` which is kind of HTTP/1.1 chunked transfer but not really. You have to send the `content-length` and include the extra bytes for framing in it, and then also set the true content-length in `x-amz-decoded-content-length`. https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html / https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming-trailers.html

I think `gun:trailers/3` makes sense to have regardless, but it looks like I won't be using it myself directly.

(Surely someone must know why S3 does this weird content encoding thing. I'll go complain :slightly_smiling_face:)

Update: it's a purely historical artifact, they invented `aws-chunked` because HTTP/1.1 chunked transfer+trailers was very immature in HTTP clients at the time. There's really no technical reason you couldn't use standard HTTP/1.1 though, if they allowed it on their end.

essen:
Yes I think it will be good to have in the next release, whenever it happens. I am working on a large H3/WT update for Cowboy that will also be done in Gun and will include other PRs with that.

PR #360, opened 2025-12-12
Add _build/ to .gitignore

meox:


essen:
Erlang.mk does not generate a `_build` directory.

PR #369, opened 2026-06-16
Allow validated control over permessage-deflate

lsund:


essen:
I obviously can't merge this. Please open a ticket requesting the feature and specifically your exact use case.

PR #373, opened 2026-07-21
Reactivate socket  when chunk response exhausts flow window

nickva:
Previously, with a flow window the final body message in a response can exhaust the window just when the stream ended. The socket then stayed passive. Any request issued next on the keepalive connection never gets its response read.

To fix it, make `send_data` emit `{active, (Flow > 0) orelse (IsFin =:= fin)}` so the message ending a stream reactivates the socket regardless of the flow window. End of stream call sites used to ignore send_data's active command (the content-length path replaced it with a hand-written {active, true}) but they now propagate it instead. The trailers path ends without a send_data call, so it reactivates the socket explicitly.

For tests use many sequential requests with flow > 1 on one connection against an origin streaming chunked bodies, with and without trailers.

### DIFF
--- a/src/gun_http.erl
+++ b/src/gun_http.erl
@@ -208,15 +208,15 @@ handle(Data, State=#http_state{in=body_chunked, in_state=InState, buffer=Buffer,
 					{fin, EvHandlerState1}
 			end,
 			%% I suppose it doesn't hurt to append an empty binary.
-			%% We ignore the active command because the stream ended.
-			[{state, State1}|_] = send_data(<<>>, State, IsFin),
+			[{state, State1}, ActiveCommand] = send_data(<<>>, State, IsFin),
 			case {HasTrailers, Conn} of
 				{trailers, _} ->
 					handle(Rest, State1#http_state{buffer = <<>>, in=body_trailer},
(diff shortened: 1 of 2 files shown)


essen:
Probably only need to change send_data's command to `{active, (Flow > 0) orelse (IsFin =:= fin)}`.

nickva:
Thanks for taking a look, Loïc. That looks like a better approach, it seems nicer to have the logic in one place. I updated the PR, hopefully I caught and updated all the call sites.

```
% gmake ct-flow
 GEN    test-build
 GEN

Common Test v1.27.7.1 starting (cwd is /Users/...)

CWD set to: "/Users/.../..."

TEST INFO: 1 test(s), 12 case(s) in 1 suite(s)

Testing src.gun.flow_SUITE: Starting test, 12 test cases
Testing src.gun.flow_SUITE: TEST COMPLETE, 12 ok, 0 failed of 12 test cases
```

essen:
Pushed a different version of the commit which I think makes more sense than previous approaches. Will be in the coming release. Closing, thanks!

nickva:
Thank you, Loïc!

